PR #7, opened 2019-05-28
Convert to go modules and test on github workflow

dgodd:
I find go modules very convenient, so I have added those files to the repo.

Seperately, to make sure that the changes I made are acceptable, and added github workflow tests (and lint)

I'm open to splitting these two apart if so required.


dgodd:
I'm completely OK with the Apache 2.0 license. Thanks for a great project

Issue #14, opened 2019-07-18
Served wrapped PAC file to point at alpaca


camh-:
Serve a "copy" of the PAC file that alapca is using that is wrapped with:

```
function FindProxyForURL(url, host) {
  return FindProxyForURL(url, host) === "DIRECT" ? "DIRECT" : "PROXY localhost:{{.Port}}";
{{.PAC}}
}
```
where `{{.Port}}` is the port that alpaca is listening on and `{{.PAC}}` is the PAC file that alpaca has fetched and is using.

This can be served on `/alpaca.pac` by alpaca.

This allows you to configure your system to use `http://localhost:3128/alpaca.pac` as the PAC file and know what URLs should be not proxied (DIRECT). All others are sent to alpaca which runs the unwrapped PAC file to proxy the request to the right place.

If you were to direct all HTTP(s) requests to alpaca in the system proxy settings, the system would use the proxy protocol always, even if alpaca is just going to go DIRECT itself. Some programs don't seem to like that they go through a proxy, so this feature would allow them to continue to go direct to their URL target.

On a Mac, one specific use case that this feature resolves is Citrix remote desktop not working when you proxy all connections via alpaca, but does work when having the system use a PAC file that does the wrapping I've provided above.

Serving files and proxy requests on the same HTTP port is possible because proxy requests all use the absolute-form (https://tools.ietf.org/html/rfc7230#section-5.3.2) and resource requests to an origin server use the origin-form (https://tools.ietf.org/html/rfc7230#section-5.3.1). alpaca can differentiate between a proxy request that it should forward and a request for a resource it serves by looking at the request-target (https://tools.ietf.org/html/rfc7230#section-5.3) and determining the form.

This would also allow other features (to be raised as separate issues if desired) such as serving the raw pac file itself (on `/proxy.pac`) that alpaca is using and allowing custom wrapping of PAC files by supplying templates similar to above by users who wish to customise the PAC routing (on `/wrapped.pac`).

PR #15, opened 2019-08-22
Serve /alpaca.pac to send non-DIRECT requests to alpaca

camh-anz:
Serve a PAC file on `/alpaca.pac` that only returns `DIRECT` or `PROXY localhost:<listen-port>` (usually `3128`). If the PAC file that alpaca is using would return `DIRECT` for a URL, the wrapped PAC also returns `DIRECT`. For everything else, it returns a proxy directive to send traffic to alpaca. If alpaca does not have a PAC file (wrong network, etc), then the PAC function only returns `DIRECT`.

This allows the system (Mac OSX, Gnome, etc) to use alpaca as the proxy, and to configure alpaca with the network proxy URL. When the system evaluates a URL with the wrapped PAC file and gets a DIRECT response, it will not proxy through alpaca and go direct instead. This helps with the odd application that doesn't seem to handle going through a proxy at all. It also means that when you're off the network that serves the PAC file, no requests go through alpaca. This makes alpaca work well as a whole-system proxy instead of just for CLI apps as originally intended.

A bit of refactoring and lint fixes made it into this PR. The most significant change would be the use of http handler wrappers (also known in some circles as http middleware). Once the code went from 1 to >1 handlers, it made sense to factor out some functionality into a handler wrapper. Some additional functionality was added that way too.

Please review this commit-by-commit and read the commit messages. I hope that makes it clearer to understand the changes and their reason. I'm happy to entertain the most nit-picky requests/reviews, so don't hold back, no matter how trivial it seems.

Fixes: #14

### DIFF
--- a/contextid.go
+++ b/contextid.go
@@ -1,0 +1,22 @@
+package main
+
+import (
+	"context"
+	"net/http"
+)
+
+// AddContextID wraps a http.Handler to add a monotonically increasing
+// uint to the context of the http.Request with the key "id" (string)
+// as it passes through the request to the next handler.
+func AddContextID(next http.Handler) http.Handler {
+	ids := make(chan uint)
+	go func() {
+		for id := uint(0); ; id++ {
+			ids <- id
+		}
+	}()
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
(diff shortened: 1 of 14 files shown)


camh-anz:
Sorry. I created this with my wrong github account. Will close and re-create.

PR #16, opened 2019-08-22
Serve /alpaca.pac to send non-DIRECT requests to alpaca

camh-:
Serve a PAC file on `/alpaca.pac` that only returns `DIRECT` or `PROXY localhost:<listen-port>` (usually `3128`). If the PAC file that alpaca is using would return `DIRECT` for a URL, the wrapped PAC also returns `DIRECT`. For everything else, it returns a proxy directive to send traffic to alpaca. If alpaca does not have a PAC file (wrong network, etc), then the PAC function only returns `DIRECT`.

This allows the system (Mac OSX, Gnome, etc) to use alpaca as the proxy, and to configure alpaca with the network proxy URL. When the system evaluates a URL with the wrapped PAC file and gets a DIRECT response, it will not proxy through alpaca and go direct instead. This helps with the odd application that doesn't seem to handle going through a proxy at all. It also means that when you're off the network that serves the PAC file, no requests go through alpaca. This makes alpaca work well as a whole-system proxy instead of just for CLI apps as originally intended.

A bit of refactoring and lint fixes made it into this PR. The most significant change would be the use of http handler wrappers (also known in some circles as http middleware). Once the code went from 1 to >1 handlers, it made sense to factor out some functionality into a handler wrapper. Some additional functionality was added that way too.

Please review this commit-by-commit and read the commit messages. I hope that makes it clearer to understand the changes and their reason. I'm happy to entertain the most nit-picky requests/reviews, so don't hold back, no matter how trivial it seems.

Fixes: #14

### DIFF
--- a/contextid.go
+++ b/contextid.go
@@ -1,0 +1,24 @@
+package main
+
+import (
+	"context"
+	"net/http"
+)
+
+// AddContextID wraps a http.Handler to add a strictly increasing
+// uint to the context of the http.Request with the key "id" (string)
+// as it passes through the request to the next handler.
+func AddContextID(next http.Handler) http.Handler {
+	// TODO(#17): Use sync/atomic AddUint64 instead of channel/goroutine
+	ids := make(chan uint)
+	go func() {
+		for id := uint(0); ; id++ {
+			ids <- id
+		}
+	}()
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		// TODO(#17): Use package scoped type instead of string for key
+		ctx := context.WithValue(req.Context(), "id", <-ids)
+		next.ServeHTTP(w, req.WithContext(ctx))
+	})
+}
--- a/contextid_test.go
+++ b/contextid_test.go
@@ -1,0 +1,35 @@
+package main
+
+import (
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func getIDFromRequest(t *testing.T, server *httptest.Server) uint {
+	res, err := http.Get(server.URL)
+	require.NoError(t, err)
+	b, err := ioutil.ReadAll(res.Body)
+	require.NoError(t, err)
+	id, err := strconv.ParseUint(string(b), 10, 64)
+	require.NoError(t, err)
+	return uint(id)
+}
+
+func TestContextID(t *testing.T) {
+	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		id, ok := r.Context().Value("id").(uint)
+		assert.True(t, ok, "Unexpected type for context id value")
+		_, err := w.Write([]byte(strconv.FormatUint(uint64(id), 10)))
+		require.NoError(t, err)
+	})
+	server := httptest.NewServer(AddContextID(handler))
+	defer server.Close()
+	assert.Equal(t, uint(0), getIDFromRequest(t, server))
+	assert.Equal(t, uint(1), getIDFromRequest(t, server))
+}
--- a/pacwrapper.go
+++ b/pacwrapper.go
@@ -1,0 +1,73 @@
+package main
+
+import (
+	"bytes"
+	"log"
+	"net/http"
+	"text/template"
+)
+
+// PACData contains program configuration to be made available to the pacWrapTmpl.
+type PACData struct {
+	Port     int
+}
+
+type pacData struct {
+	PACData
+	UpstreamPAC string
+}
+
+type PACWrapper struct {
+	data      pacData
+	tmpl      *template.Template
+	alpacaPAC string
+}
+
+// PACWrapper template for serving a PAC file to point at alpaca or DIRECT. If we have a valid
+// PAC file, we wrap that PAC file with a wrapper function that only returns "DIRECT" or
+// "localhost:port". If we do not have a PAC file, the PAC function we serve only returns "DIRECT",
+// which should prevent all requests reaching us.
+var pacWrapTmpl = `// Wrapped for and by alpaca
+function FindProxyForURL(url, host) {
+{{ if .UpstreamPAC }}
+  return FindProxyForURL(url, host) === "DIRECT" ? "DIRECT" : "PROXY localhost:{{.Port}}";
+{{.UpstreamPAC}}
+{{ else }}
+  return "DIRECT";
+{{ end }}
+}
+`
+
+func NewPACWrapper(data PACData) *PACWrapper {
+	t := template.Must(template.New("alpaca").Parse(pacWrapTmpl))
+	return &PACWrapper{pacData{data, ""}, t, ""}
+}
+
+func (pw *PACWrapper) Wrap(pacjs []byte) {
+	pac := string(pacjs)
+	if pac == pw.data.UpstreamPAC && pw.alpacaPAC != "" {
+		return
+	}
+	pw.data.UpstreamPAC = pac
+	b := &bytes.Buffer{}
+	if err := pw.tmpl.Execute(b, pw.data); err != nil {
+		log.Printf("error executing PAC wrap template: %v\n", err)
+		return
+	}
+	pw.alpacaPAC = b.String()
+}
+
+func (pw *PACWrapper) SetupHandlers(mux *http.ServeMux) {
+	mux.HandleFunc("/alpaca.pac", pw.handlePAC)
+}
+
+func (pw *PACWrapper) handlePAC(w http.ResponseWriter, req *http.Request) {
+	if req.Method != http.MethodGet {
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
+	w.Header().Set("Content-Type", "application/x-ns-proxy-autoconfig")
+	if _, err := w.Write([]byte(pw.alpacaPAC)); err != nil {
+		log.Printf("Error writing PAC to response: %v\n", err)
+	}
+}
--- a/pacwrapper_test.go
+++ b/pacwrapper_test.go
@@ -1,0 +1,47 @@
+package main
+
+import (
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestWrapPAC(t *testing.T) {
+	pw := NewPACWrapper(PACData{Port: 1234})
+	pac := `function FindProxyForURL(url, host) { return "DIRECT" }`
+	pw.Wrap([]byte(pac))
+	assert.Contains(t, pw.alpacaPAC, pac)
+	assert.Contains(t, pw.alpacaPAC, `"DIRECT" : "PROXY localhost:1234"`)
+}
+
+func TestWrapEmptyPAC(t *testing.T) {
+	pw := NewPACWrapper(PACData{Port: 1234})
+	pw.Wrap(nil)
+	assert.Contains(t, pw.alpacaPAC, `return "DIRECT"`)
+}
+
+func TestPACServe(t *testing.T) {
+	pw := NewPACWrapper(PACData{Port: 1234})
+	pac := `function FindProxyForURL(url, host) { return "DIRECT" }`
+	pw.Wrap([]byte(pac))
+	mux := http.NewServeMux()
+	pw.SetupHandlers(mux)
+	server := httptest.NewServer(mux)
+	defer server.Close()
+
+	resp, err := http.Get(server.URL + "/alpaca.pac")
+	require.NoError(t, err)
+
+	assert.Equal(t, resp.StatusCode, http.StatusOK)
+	assert.Equal(t, "application/x-ns-proxy-autoconfig", resp.Header.Get("Content-Type"))
+	b, err := ioutil.ReadAll(resp.Body)
+	body := string(b)
+	require.NoError(t, err)
+	assert.Contains(t, body, pac)
+	assert.Contains(t, body, `"DIRECT" : "PROXY localhost:1234"`)
+	resp.Body.Close()
+}
--- a/proxy_test.go
+++ b/proxy_test.go
@@ -5,8 +5,6 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 	"io"
 	"io/ioutil"
 	"net"
@@ -14,6 +12,9 @@ import (
 	"net/http/httptest"
 	"net/url"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 type testServer struct {
@@ -74,7 +75,9 @@ func TestGetViaProxy(t *testing.T) {
 	requests := make(chan string, 2)
 	server := httptest.NewServer(testServer{requests})
 	defer server.Close()
-	proxy := httptest.NewServer(testProxy{requests, "proxy", newDirectProxy()})
+	// Proxy request should not go to the mux. The empty mux will always return 404.
+	mux := http.NewServeMux()
+	proxy := httptest.NewServer(testProxy{requests, "proxy", newDirectProxy().WrapHandler(mux)})
 	defer proxy.Close()
 	tr := &http.Transport{Proxy: proxyServer(t, proxy)}
 	testGetRequest(t, tr, server.URL)
@@ -87,13 +90,29 @@ func TestGetOverTlsViaProxy(t *testing.T) {
 	requests := make(chan string, 2)
 	server := httptest.NewTLSServer(testServer{requests})
 	defer server.Close()
-	proxy := httptest.NewServer(testProxy{requests, "proxy", newDirectProxy()})
+	// Proxy request should not go to the mux. The empty mux will always return 404.
+	mux := http.NewServeMux()
+	proxy := httptest.NewServer(testProxy{requests, "proxy", newDirectProxy().WrapHandler(mux)})
 	defer proxy.Close()
 	tr := &http.Transport{Proxy: proxyServer(t, proxy), TLSClientConfig: tlsConfig(server)}
 	testGetRequest(t, tr, server.URL)
 	require.Len(t, requests, 2)
 	assert.Equal(t, "CONNECT to proxy", <-requests)
 	assert.Equal(t, "GET to server", <-requests)
+}
+
+func TestGetOriginURLsNotProxied(t *testing.T) {
+	requests := make(chan string, 2)
+	mux := http.NewServeMux()
+	mux.HandleFunc("/origin", func(w http.ResponseWriter, req *http.Request) {
+		_, err := w.Write([]byte("Hello, client\n"))
+		require.NoError(t, err)
+	})
+	proxy := httptest.NewServer(testProxy{requests, "proxy", newDirectProxy().WrapHandler(mux)})
+	defer proxy.Close()
+	testGetRequest(t, &http.Transport{}, proxy.URL+"/origin")
+	require.Len(t, requests, 1)
+	assert.Equal(t, "GET to proxy", <-requests)
 }
 
 func TestGetViaTwoProxies(t *testing.T) {
--- a/proxyfinder.go
+++ b/proxyfinder.go
@@ -14,11 +14,12 @@ import (
 type ProxyFinder struct {
 	runner  *PACRunner
 	fetcher *pacFetcher
+	wrapper *PACWrapper
 	sync.Mutex
 }
 
-func NewProxyFinder(pacurl string) *ProxyFinder {
-	pf := &ProxyFinder{}
+func NewProxyFinder(pacurl string, wrapper *PACWrapper) *ProxyFinder {
+	pf := &ProxyFinder{wrapper: wrapper}
 	if len(pacurl) == 0 {
 		log.Println("No PAC URL specified or detected; all requests will be made directly")
 	} else if _, err := url.Parse(pacurl); err != nil {
@@ -48,10 +49,15 @@ func (pf *ProxyFinder) checkForUpdates() {
 	var pacjs []byte
 	pacjs = pf.fetcher.download()
 	if pacjs == nil {
+		if !pf.fetcher.isConnected() {
+			pf.wrapper.Wrap(nil)
+		}
 		return
 	}
 	if err := pf.runner.Update(pacjs); err != nil {
 		log.Printf("Error running PAC JS: %q\n", err)
+	} else {
+		pf.wrapper.Wrap(pacjs)
 	}
 }
 
--- a/proxyfinder.go
+++ b/proxyfinder.go
@@ -31,6 +31,17 @@ func NewProxyFinder(pacurl string) *ProxyFinder {
 	return pf
 }
 
+func (pf *ProxyFinder) WrapHandler(next http.Handler) http.Handler {
+	// If we don't have a fetcher, don't wrap the handler as there's nothing to do.
+	if pf.fetcher == nil {
+		return next
+	}
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		pf.checkForUpdates()
+		next.ServeHTTP(w, req)
+	})
+}
+
 func (pf *ProxyFinder) checkForUpdates() {
 	pf.Lock()
 	defer pf.Unlock()
@@ -50,7 +61,6 @@ func (pf *ProxyFinder) findProxyForRequest(req *http.Request) (*url.URL, error) 
 		log.Printf(`[%d] %s %s via "DIRECT"`, id, req.Method, req.URL)
 		return nil, nil
 	}
-	pf.checkForUpdates()
 	if !pf.fetcher.isConnected() {
 		log.Printf(`[%d] %s %s via "DIRECT" (not connected)`, id, req.Method, req.URL)
 		return nil, nil
--- a/proxyfinder.go
+++ b/proxyfinder.go
@@ -18,7 +18,15 @@ type ProxyFinder struct {
 }
 
 func NewProxyFinder(pacurl string) *ProxyFinder {
-	pf := &ProxyFinder{runner: new(PACRunner), fetcher: newPACFetcher(pacurl)}
+	pf := &ProxyFinder{}
+	if len(pacurl) == 0 {
+		log.Println("No PAC URL specified or detected; all requests will be made directly")
+	} else if _, err := url.Parse(pacurl); err != nil {
+		log.Fatalf("Couldn't find a valid PAC URL: %v", pacurl)
+	} else {
+		pf.runner = new(PACRunner)
+		pf.fetcher = newPACFetcher(pacurl)
+	}
 	pf.checkForUpdates()
 	return pf
 }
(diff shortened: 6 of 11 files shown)


juliaogris:
@samuong - you should protect your `master` branch

pacfetcher_test.go
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 func pacjsHandler(pacjs string) http.HandlerFunc {
-	return func(w http.ResponseWriter, r *http.Request) { w.Write([]byte(pacjs)) }
+	return func(w http.ResponseWriter, r *http.Request) { w.Write([]byte(pacjs)) } // nolint:errcheck
```
juliaogris: I think `_, _ := w.Write(...)` also gets rid of the linter warning (same for all the others below).
Did you intentionally choose the `// nolint: ...` ?

contextid.go
```diff
@@ -0,0 +1,22 @@
+package main
+
+import (
+	"context"
+	"net/http"
+)
+
+// AddContextID wraps a http.Handler to add a monotonically increasing
```
juliaogris: I don't think "monotonically" is the right word here:
>  monotonic increasing: (mathematics, of a function) always increasing or remaining constant, and never decreasing; contrast this with strictly increasing ([wiktionary](https://en.wiktionary.org/wiki/monotonic_increasing))
IDs should be strictly increasing, right?

contextid.go
```diff
@@ -0,0 +1,22 @@
...
+// as it passes through the request to the next handler.
+func AddContextID(next http.Handler) http.Handler {
+	ids := make(chan uint)
+	go func() {
+		for id := uint(0); ; id++ {
+			ids <- id
+		}
+	}()
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		ctx := context.WithValue(req.Context(), "id", <-ids)
```
juliaogris: I think it's recommended not to use primitive types as context keys. If some other middleware adds a value with key "id" your id gets overridden.
`context.WithValue(parentCtx, key, value)` takes `key` parameter of type `interface{}`.
Two interface values are equal if they have identical dynamic types and equal dynamic values or if both have value nil.
See https://golang.org/pkg/context/#WithValue and https://golang.org/ref/spec#Comparison_operators

contextid_test.go
```diff
@@ -0,0 +1,35 @@
...
+	assert.NoError(t, err)
+	id, err := strconv.ParseUint(string(b), 10, 64)
+	assert.NoError(t, err)
+	return uint(id)
+}
+
+func TestContextId(t *testing.T) {
+	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		id := r.Context().Value("id")
+		_, err := w.Write([]byte(strconv.FormatUint(uint64(id.(uint)), 10)))
```
juliaogris: more likely (whatever that means) that the type assertion panics than that w.Write errors I reckon.

pacfetcher_test.go
camh-: Sam said a similar thing. I just wanted to silence the warnings because i wanted to see only lint warnings from my code. The comment says I am silencing the warnings. The underscores says I am ignoring the warnings. Ignoring them should be a deliberate decision, one that I wasn't going to make. With the comment, it is easy to grep for that and revisit if these are to be fixed properly by doing something with the error. The other form is not as easy to grep for.

I'll leave it up to Sam what he wants done with this, and then tell him they're his lint warnings anyway :)

contextid.go
camh-: Done. Changed to "strictly increasing" locally. Will push the fix after further reviews come in.

contextid.go
camh-: Yes, that's correct. You could create an issue for it. But this is how the code was already - I've just moved where the id is generated and added. I don't plan to fix that (in this PR, or maybe at all).

contextid_test.go
camh-: I think you might be saying that I should test explicitly for the type of the id, and I agree. I will add that test. As to handling the `Write` error, that was to silence a lint warning (and since I wrote the code, I handled the error). Is there a common way in tests to differentiate between assertions that validate the results of the test and assertions that just make sure the test works as expected? This is the latter so I don't want to ignore the error - may as well fail the test and understand why the error occurred, but it might be nice for the developer to know that the error is in the test case and not the unit under test.

camh-: I've split out the type assertion and asserted that it is ok. So no panic now, but it is tested that the type is correct. I didn't make a new test for this. New code coming with the next push.

proxyfinder.go
```diff
@@ -31,6 +31,15 @@ func NewProxyFinder(pacurl string) *ProxyFinder {
 	return pf
 }
 
+func (pf *ProxyFinder) WrapHandler(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		if pf.fetcher != nil {
```
anzdaddy: Does `pf.fetcher` change after starting? If not, move this check outside `return http.HandlerFunc(…)`.

If it's the case in practice, but not guaranteed, consider guaranteeing it (or raising an issue to address this later).

pacwrapper.go
```diff
@@ -0,0 +1,85 @@
...
+	mux.HandleFunc("/alpaca.pac", pw.handleAlpacaPAC)
+}
+
+func (pw *PACWrapper) handleProxyPAC(w http.ResponseWriter, req *http.Request) {
+	pw.handlePAC(w, req, pw.data.PAC)
+}
+
+func (pw *PACWrapper) handleAlpacaPAC(w http.ResponseWriter, req *http.Request) {
+	pw.handlePAC(w, req, pw.alpacaPAC)
+}
```
anzdaddy: nit: I'd be sorely tempted to inline these in `SetupHandlers`.

pacwrapper_test.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestWrapPAC(t *testing.T) {
```
anzdaddy: `t.Parallel()` at the front of all tests where practicable.

pacfetcher_test.go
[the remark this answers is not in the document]
camh-: That link says the Go convention is to not put a space after the comment marker for machine readable comments. That is clearly wrong because godoc comments are machine readable and always have the space, but even build switches in Go do not have it: https://github.com/golang/go/blob/9f78aa77443389d61a3ef910a1c3d4628999d82c/src/runtime/mmap.go#L5

I might be more willing to accept what the golangci lint docs say if they provided a reference or source for their assertion, but without it all I can say is that I have never seen the no-space comment thing anywhere.

I'm going to stick to the much nicer slash-slash-space style until I see something otherwise.

contextid_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"net/http/httptest"
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func getIdFromRequest(t *testing.T, server *httptest.Server) uint {
+	res, err := http.Get(server.URL)
+	assert.NoError(t, err)
```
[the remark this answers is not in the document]
camh-: I like the idea of using `require` in test cases where an error indicates the test itself is broken rather than the unit under test. I will re-work my tests to use `require` in those circumstances.

contextid_test.go
```diff
@@ -0,0 +1,35 @@
...
+		id := r.Context().Value("id")
+		_, err := w.Write([]byte(strconv.FormatUint(uint64(id.(uint)), 10)))
+		assert.NoError(t, err)
+	})
+	server := httptest.NewServer(AddContextID(handler))
+	defer server.Close()
+	id0 := getIdFromRequest(t, server)
+	assert.Equal(t, id0, uint(0))
+	id1 := getIdFromRequest(t, server)
+	assert.Equal(t, id1, uint(1))
```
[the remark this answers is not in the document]
camh-: That is reasonable. I will do that.

pacwrapper.go
camh-: I was tempted and I also tried another approach of type aliasing `string` to `pacString` and defining a `ServeHTTP` method on it, so as to remove these two single line functions. But the anonymous function I felt was a little too verbose in the function signature (`w http.ResponseWriter, req *http.Request)` is a little to long with the func and embedded in the function call), and the type alias didn't work unless it was a pointer which had an impact on `text/template` use of that.

This does feel like a little bit of a wart but that seems to happen somewhat often with Go. It is quite readable as is, so I'll stick with it unless there's something I find definitively better.

pacwrapper_test.go
camh-: I agree that's a useful thing, but I'm not going to start a few tests like that and leave the rest alone. That is something that should be considered across the whole code base I think, and is out of scope for this PR.

proxyfinder.go
```diff
@@ -18,7 +18,15 @@ type ProxyFinder struct {
...
 
 func NewProxyFinder(pacurl string) *ProxyFinder {
-	pf := &ProxyFinder{runner: new(PACRunner), fetcher: newPACFetcher(pacurl)}
+	pf := &ProxyFinder{}
+	if len(pacurl) == 0 {
+		log.Println("No PAC URL specified or detected; all requests will be made directly")
+	} else if _, err := url.Parse(pacurl); err != nil {
+		log.Fatalf("Couldn't find a valid PAC URL: %v", pacurl)
+	} else {
+		pf.runner = new(PACRunner)
```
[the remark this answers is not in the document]
camh-: yep. I noticed that, but that commit is a refactor of the structure, so I wanted to keep the rest the same. But as I considered it, I could not really come up with a reason for changing `new(X)` to `X{}`, when you often have custom constructors like `NewX()`. There are too many ways to create a new thing in Go.

contextid_test.go
camh-: > strong guarantee about what's coming out

Well, only if the code is correct. And the purpose of these tests is to confirm that. I think an assertion is still the correct form of detection of incorrect code.

pacfetcher_test.go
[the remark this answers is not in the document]
anzdaddy: go generate requires it. https://blog.golang.org/generate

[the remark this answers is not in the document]
camh-: I started arguing, but bah - I don't care enough.

Done

contextid.go
```diff
@@ -0,0 +1,22 @@
...
+// AddContextID wraps a http.Handler to add a monotonically increasing
+// uint to the context of the http.Request with the key "id" (string)
+// as it passes through the request to the next handler.
+func AddContextID(next http.Handler) http.Handler {
+	ids := make(chan uint)
+	go func() {
+		for id := uint(0); ; id++ {
+			ids <- id
+		}
+	}()
```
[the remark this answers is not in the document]
camh-: Added to #17 - trivial enough not to need its own issue. Not fixing now because I just refactored what was there - fixing it is outside the scope of this PR.

proxyfinder.go
camh-: No it does not change. So I moved the check outside the HandlerFunc.

Done.

contextid.go
camh-: Fixed in 8ecf68b7141491dcacaa83cc64922fb73463530a. Will rebase-squash prior to merging.

contextid_test.go
camh-: Fixed in 9c865583b25d7fd8d8c4315c2ccd2cbe7ae90155. Will rebase-squash prior to merging.

pacfetcher_test.go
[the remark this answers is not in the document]
camh-: Fixed in fc7715bf8dbb439e996234d0949c385ac6b3ce0c and 0fac3dda2d8e3bfc6c9c2d802b37cc4fa9ac560c. Will rebase-squash prior to merging.

contextid_test.go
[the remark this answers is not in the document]
camh-: Fixed in 719352970088b7beb366fa5b4c354ba42e672213, 9ddd9a5c4026641553706ace54060736c6f13150 and b174e824ff71a4e78e82f937477c7ad426b59f91. Will rebase-squash prior to merging.

contextid_test.go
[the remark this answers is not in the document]
camh-: Fixed in 50e4d585f2b313f3b4a3f6b9b26ec4c5f9ff8110. Will rebase-squash prior to merging.

proxyfinder.go
camh-: Fixed in 82a1dd29ef1b88e614aa5e9c8559680c85401435. Will rebase-squash prior to merging.

pacfetcher_test.go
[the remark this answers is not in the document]
camh-: Fixed in fc7715bf8dbb439e996234d0949c385ac6b3ce0c and 0fac3dda2d8e3bfc6c9c2d802b37cc4fa9ac560c. Will rebase-squash prior to merging.

camh-:
@juliaogris @anzdaddy Changes pushed as fixup commits. PTAL, and resolve all comments that you feel are now resolved. Or continue to argue as you see fit.

samuong:
> @samuong - IMO you should protect and "require pull request reviews before merging" for your `master` branch. Maybe ;)

@juliaogris The only reason I haven't done that in the past is that I didn't have anyone else to review my own pull requests. You all are the only ones (other than me) who have commented on PRs, so if you're happy to be collaborators and review my code, then I'll add you to the repo.

pacfetcher_test.go
samuong: If you're going to start using a linter, can you document this in something like CONTRIBUTING.md? I'll go through the lint errors in my code and fix them.

juliaogris: Better also covered CI - Github actions? Travis?  Beyond the scope of this PR I'd say.
I've created an issue and assigned to myself #20 (including creation of CONTRIBUTING.md)

juliaogris: Also @anzdaddy I don't think `golangci-lint 1.17` with "strict" settings complains about `_, _ := ...`.
"strict" being:
```yaml
issues:
  exclude-use-default: false
linters:
  enable-all: true
```

contextid.go
juliaogris: actually I still think this is not the right term (sorry should have noticed earlier).
It's only used for continuous and not discrete functions, but I might be wrong here...
Doesn't matter all that much, you communicate clearly what it is: an "autoincrement", "serial" kinda thing from the DB perspective.

pacwrapper_test.go
juliaogris: nice one @anzdaddy - I did not know about this (agreed with @camh- should be part of a separate refactor).
I wonder if it's worth while using suites and setting this kind of thing in `SetupTest`?

contextid_test.go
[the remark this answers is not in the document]
juliaogris: bit late to the party, sorry. An alternative that is a bit more explicitly (if you don't know the difference between testify's `require` and `assert` off the top of your head) and to be used if needed rarely is: if testbroken { [assert.FailNow()](https://godoc.org/github.com/stretchr/testify/assert#FailNow)  }`  

Aside: you can force require behaviour for _all_ assert statements with `go test -failfast` I believe.

requestlogger.go
```diff
@@ -0,0 +1,24 @@
...
+func (w *statusWriter) WriteHeader(status int) {
+	w.status = status
+	w.ResponseWriter.WriteHeader(status)
+}
+
+func RequestLogger(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		sw := &statusWriter{ResponseWriter: w, status: http.StatusOK}
+		next.ServeHTTP(sw, req)
+		log.Printf("[%v] %d %s %s\n", req.Context().Value("id"), sw.status, req.Method, req.URL)
```
juliaogris: would elapsed time be interesting here too?

pacfetcher_test.go
marcelocantos:
```yaml
linters-settings:
  errcheck:
    check-blank: true
```

contextid.go
marcelocantos: The definition “for all _x_ and _y_ such that _x_ < _y_ one has _f_(_x_) < _f_(_y_)” applies equally well to discrete functions.

juliaogris:
Kawoom! :)

contextid_test.go
```diff
@@ -22,8 +22,9 @@ func getIdFromRequest(t *testing.T, server *httptest.Server) uint {
 
 func TestContextId(t *testing.T) {
 	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		id := r.Context().Value("id")
-		_, err := w.Write([]byte(strconv.FormatUint(uint64(id.(uint)), 10)))
+		id, ok := (r.Context().Value("id")).(uint)
```
marcelocantos: nit: Redundant parentheses.

pacfetcher_test.go
juliaogris: 💥  :)

contextid.go
juliaogris: 💥  :)

requestlogger.go
camh-: I think there's a lot of useful things that could be logged. Sam and I talked about a logging refactor to use a standardised log format for all logs. For now, I'm just aiming for consistency with the rest of the code.

contextid_test.go
camh-: Done.

Fixed in 6f754210a8dcc37c5f164656f75c22a5a6c8449d

pacfetcher_test.go
camh-: The linter settings to use should be discussed in a relevant issue, not here. It will get lost here.

@samuong I will remove the //nolint commit from this PR and re-work my code where I added one so the topic of linting disappears from this PR, and it can be covered in #20 (Setup CI) and your PR. To do so, I'll squash all my fixes so far into the relevant commits and force push the branch with the reduced set of commits.

contextid.go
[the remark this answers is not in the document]
camh-: Added TODO to code.

camh-:
@marcelocantos @juliaogris @samuong All comments resolved. I have forced-pushed my squashed series, so back to just the commits that matter without fixups. I have removed the `//nolint` directives and added a couple of TODOs.

PTAL

contextid_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"net/http"
+	"net/http/httptest"
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func getIdFromRequest(t *testing.T, server *httptest.Server) uint {
```
samuong: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

contextid_test.go
```diff
@@ -0,0 +1,37 @@
...
+	res, err := http.Get(server.URL)
+	require.NoError(t, err)
+	b, err := ioutil.ReadAll(res.Body)
+	require.NoError(t, err)
+	id, err := strconv.ParseUint(string(b), 10, 64)
+	require.NoError(t, err)
+	return uint(id)
+}
+
+func TestContextId(t *testing.T) {
```
samuong: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

contextid_test.go
```diff
@@ -0,0 +1,37 @@
...
+		assert.True(t, ok, "Unexpected type for context id value")
+		_, err := w.Write([]byte(strconv.FormatUint(uint64(id), 10)))
+		require.NoError(t, err)
+	})
+	server := httptest.NewServer(AddContextID(handler))
+	defer server.Close()
+	id0 := getIdFromRequest(t, server)
+	id1 := getIdFromRequest(t, server)
+	assert.True(t, uint(0) <= id0)
+	assert.True(t, id0 < id1)
```
samuong: I'm wondering if there's any reason you're not just checking for 0 and 1 here - is it just because the doc comment says "strictly increasing", and you don't want to test anything more specific than that?

I remember reading a discussion earlier about "monotonically increasing" vs "strictly increasing", but wouldn't things be simpler if you just used the word "sequential"?

proxy_test.go
```diff
@@ -74,7 +75,9 @@ func TestGetViaProxy(t *testing.T) {
 	requests := make(chan string, 2)
 	server := httptest.NewServer(testServer{requests})
 	defer server.Close()
-	proxy := httptest.NewServer(testProxy{requests, "proxy", newDirectProxy()})
+	// Proxy request should not go to the mux. The empty mux will always return 404
+	mux := http.NewServeMux()
+	proxy := httptest.NewServer(testProxy{requests, "proxy", newDirectProxy().WrapHandler(mux)})
```
samuong: perfect line length 💯

proxy_test.go
```diff
@@ -74,7 +75,9 @@ func TestGetViaProxy(t *testing.T) {
 	requests := make(chan string, 2)
 	server := httptest.NewServer(testServer{requests})
 	defer server.Close()
-	proxy := httptest.NewServer(testProxy{requests, "proxy", newDirectProxy()})
+	// Proxy request should not go to the mux. The empty mux will always return 404
```
samuong: Nit: need a period after the end of the sentence.

pacwrapper.go
```diff
@@ -0,0 +1,85 @@
...
+	"log"
+	"net/http"
+	"text/template"
+)
+
+type PACData struct {
+	Port     int
+	PACURL   string
+	Domain   string
+	Username string
```
samuong: What does this use the Domain and Username for? I couldn't find any references to it elsewhere in the code...

samuong: ...also I couldn't figure out what PACURL is used for either...

pacwrapper.go
```diff
@@ -0,0 +1,85 @@
...
+	b := &bytes.Buffer{}
+	if err := pw.tmpl.Execute(b, pw.data); err != nil {
+		log.Printf("error executing PAC wrap template: %v\n", err)
+		return
+	}
+	pw.alpacaPAC = b.String()
+}
+
+func (pw *PACWrapper) SetupHandlers(mux *http.ServeMux) {
+	mux.HandleFunc("/proxy.pac", pw.handleProxyPAC)
```
samuong: I think we discussed this before and I still can't think of a real use case for serving the upstream PAC file again. It feels like something I wouldn't want people to rely on (in case we want to make a breaking change in the future) so I'd prefer to get rid of this.

pacwrapper.go
```diff
@@ -0,0 +1,85 @@
...
+type PACData struct {
+	Port     int
+	PACURL   string
+	Domain   string
+	Username string
+}
+
+type pacData struct {
+	PACData
+	PAC string
```
samuong: The name "PAC" feels a little ambiguous to me, can we make it something more descriptive? How about UpstreamPAC?

pacwrapper.go
```diff
@@ -0,0 +1,85 @@
...
+	data      pacData
+	tmpl      *template.Template
+	alpacaPAC string
+}
+
+// PACWrapper template for serving a PAC file to point at alpaca or DIRECT.
+// If we have a valid PAC file, we wrap that PAC file with a wrapper
+// function that only returns "DIRECT" or "localhost:port". If we do not
+// have a PAC file, the PAC function we serve only returns "DIRECT", which
+// should prevent all requests reaching us.
```
samuong: If we're going to go with a 100 char line length, I think it'd be nice to wrap all the comments at 100 chars rather than 80 or whatever this is wrapped at.

pacwrapper.go
```diff
@@ -0,0 +1,85 @@
...
+func (pw *PACWrapper) handleProxyPAC(w http.ResponseWriter, req *http.Request) {
+	pw.handlePAC(w, req, pw.data.PAC)
+}
+
+func (pw *PACWrapper) handleAlpacaPAC(w http.ResponseWriter, req *http.Request) {
+	pw.handlePAC(w, req, pw.alpacaPAC)
+}
+
+func (pw *PACWrapper) handlePAC(w http.ResponseWriter, req *http.Request, pac string) {
+	if req.Method != "GET" {
```
samuong: Let's be consistent and use http.MethodGet

contextid_test.go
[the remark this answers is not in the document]
samuong: I made a comment elsewhere about this but then I found this discussion here. What's wrong with id0==0 and id1==1? I'm not sure what we're gaining by loosening the contract...

contextid_test.go
camh-: I guess sequential means to follow a sequence (https://en.wikipedia.org/wiki/Integer_sequence) But what sequence? The negafibonacci sequence is not strictly increasing for instance.

But I don't really care. I did test originally with 0 and 1, but I updated this test responding to @anzdaddy's comment (https://github.com/samuong/alpaca/pull/16#discussion_r317423137)

Let me know if you want me to change it back.

pacwrapper.go
camh-: I had in mind the next patch where a user would be able to supply their own wrapper, so I added all the context available to the struct so it would be available to the template. I can remove it if you like. Or should I write that next patch?

contextid_test.go
camh-: Done. Fixed in 25c8a2b061eb2c75af8ea98d0cb42faa6f4605d0

contextid_test.go
camh-: Done. Fixed in 25c8a2b061eb2c75af8ea98d0cb42faa6f4605d0

contextid_test.go
anzdaddy: You could go even further and assert merely that id0 ≠ id1, since that's all that's required for correct behaviour. But we're in serious danger of overthinking this. I'll be happy with any assertion that passes when id0 = id1 - 1 = 0 and fails when id0 = id1.

proxy_test.go
camh-: Done. Fixed in e8a57324da094f3e748dc08e24f188b127c4fd5a

pacwrapper.go
camh-: Done. Fixed in 93581e296d6f7b415cdeb3150015bc53832a7faa

pacwrapper.go
camh-: Done. Fixed in eccb52048c68062948af4e90bdc3b6d295c8d47b

pacwrapper.go
camh-: I wrap comments at 80 otherwise text becomes too hard to read. Or more that I don't let comment lines become too long - if they start at column 20, then I would be ok going to 100.

I didn't know 100 char line length had been decided on though. I have reformatted to 100 chars wide for the comment.

Done. Fixed in 46688aebb07260eac05dc4424dc9e8a3540650a9

pacwrapper.go
camh-: Done. Fixed in 72d31e7ee953938c6cbfb42365e90b05313255ea

camh-:
@samuong PTAL

I've pushed fixup commits and linked them to the comments so you can see how I have addressed your comments. When you've re-reviewed, I'll run `git rebase -i --autosquash master` to squash the fixups into the relevant commits and force push so there is still a clean series.

pacwrapper.go
samuong: If it's part of another patch, I do feel like it should go together with that other patch.

contextid_test.go
samuong: The reason I asked was because I wanted to see if we could turn this into two lines of code. If you check that the ids are 0 and 1, then you could write

```
assert.True(t, uint(0), getIdFromRequest(t, server))
assert.True(t, uint(1), getIdFromRequest(t, server))
```

I guess if you just checked that id0 != id1, you could do it like this:

```
id0 := getIdFromRequest()
id1 := getIdFromRequest()
assert.NotEqual(t, id0, id1)
```

Not sure if undefined order of evaluation of arguments is a thing in Go, so maybe it doesn't work reliably in a single line? It's probably a little awkward to read with just one line though...

samuong: Also `getIdFromRequest()` returns a `uint`, so an assertion like `assert.True(t, uint(0) <= id0)` could never fail, right?

camh-: Ok. Done in 665f24cf2f84f6a9aaef096bc56f750bd4a34cc2

pacwrapper.go
camh-: Ok. Done in e928fe0a4591e0ed00b54e5e6f97fc807de6dda7

camh-:
PTAL @samuong. Again, fixup commits pushed which I'll squash into the main series when you're happy with it.

samuong:
Thanks Cam! LGTM.

I'll let you squash everything before I actually approve it, since I think your squash will invalidate my approval anyway. But once you're done I'm happy to give the actual approval.

camh-:
Ok. Squashed and force-pushed. If you click on the "force-pushed" link where it says "camh- force-pushed the camh-:pacwrap branch" you can verify the push did not alter anything.

camh-:
@samuong I wondered if this was going to happen. It looks like you're not a CODEOWNER so you cannot approve this merge. Only @juliaogris or @marcelocantos can :-)

samuong:
https://github.com/samuong/alpaca/pull/25

PR #19, opened 2019-08-28
Create CODEOWNERS

samuong:


samuong:
> Now @camh- would say: this is a fixup commit, squash it ;)

I'm wondering whether this project should only allow squash merging, in which case it'll get squashed when I merge this into master. Would be good to get your thoughts. This might require a long discussion :)

juliaogris:
Yes, that will indeed be a longer discussion.
I'll eat my hat if @camh- favours forced squash merges....

camh-:
I am honoured to be first, but please sort descending by age.

camh-:
I would be opposed to squash merging as factoring commits is not just useful for review but also for code archaeology, cherry picking, reverting and many other useful features. In these cases you would want to maintain the small commits in master and not squash them. However, I could get on board with squash merging if @juliaogris will make a video of her eating her hat.

Issue #21, opened 2019-09-11
Safari doesn't work through Alpaca


samuong:
Here's what's going on:

1. Safari sends a CONNECT request to Alpaca
2. Alpaca responds with a 200 OK, and Go's `ResponseWriter.WriteHeader()` will automatically add `Transfer-Encoding: chunked` to the response header
3. It then blindly copies data back and forth between the client and the server, but this data is _not_ chunk encoded, and Safari's chunk decoder appears to fail as a result

According to https://tools.ietf.org/html/rfc7231#section-4.3.6:

> A server MUST NOT send any Transfer-Encoding or Content-Length header fields in a 2xx (Successful) response to CONNECT.  A client MUST ignore any Content-Length or Transfer-Encoding header fields received in a successful response to CONNECT.

It's not clear to me why Safari isn't ignoring the Transfer-Encoding header from Alpaca, but I think it should.

This also needs to be fixed in Alpaca by writing the CONNECT response directly to the connection (using `Conn.Write()`), rather than using Go's `ResponseWriter`.

PR #23, opened 2019-09-18
Avoid sending Content-Length on successful CONNECT response

samuong:
Fixes: #21

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -99,6 +99,7 @@ func connectViaProxy(w http.ResponseWriter, req *http.Request, proxy string, aut
 	}
 	err = req.Write(conn)
 	if err != nil {
+		conn.Close()
 		log.Printf("[%d] Error sending CONNECT request: %v", req.Context().Value("id"), err)
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return nil
@@ -120,6 +121,7 @@ func connectViaProxy(w http.ResponseWriter, req *http.Request, proxy string, aut
 		rd = bufio.NewReader(conn)
 		resp, err = auth.connect(req, conn, rd)
 		if err != nil {
+			conn.Close()
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return nil
 		}
--- a/proxy.go
+++ b/proxy.go
@@ -76,11 +76,11 @@ func (ph ProxyHandler) handleConnect(w http.ResponseWriter, req *http.Request) {
 		server.Close()
 		return
 	}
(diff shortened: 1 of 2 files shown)


samuong:
Ugh, so after testing this a bit at work, I realise there are still another case that need fixing. Will do so soon, but maybe hold off on doing a code review unless you really want to.

samuong:
Ok, this has been working well for me today. I've tested it with Safari, browsing to both internal sites (Safari->Alpaca->website) as well as external ones (Safari->Alpaca->corporate proxy->website). I've also checked the JAMF Self Service isn't broken either (I'm able to download VSCode). PTAL?

proxy.go
```diff
@@ -51,108 +51,86 @@ func (ph ProxyHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+		server, err = connectViaProxy(req, u.Host, ph.auth)
 	}
-	if server == nil {
+	if err != nil {
+		w.WriteHeader(http.StatusBadGateway)
+		return
+	}
+	h, ok := w.(http.Hijacker)
+	if !ok {
+		msg := fmt.Sprintf("Can't hijack connection to %v", req.Host)
```
camh-: This message is a little strange. You can't hijack the connection because it is not hijackable. It doesn't really have anything to do with the `req.Host` that you want to connect to. This is more of an internal error, but the message suggests some sort of external cause. And if you are going via a proxy, the connection isn't really "to `req.Host`" either.

samuong: Yes, you're right. Technically it's a connection going back to `req.RemoteAddr`. If I changed it to say that, do you think that's enough? Or are you also saying that it's just not a very user friendly error message?

samuong: Also it's not really hijacking a connection, it's hijacking a response (writer). So how about the following?

`fmt.Sprintf("Can't hijack response back to %v", req.RemoteAddr)`

camh-: I guess it doesn't really matter what it says - there's not much you can do if you get the error - I'm not sure using `req.RemoteAddr` makes much difference either. The original message just seemed odd when I read it, but I guess `req.Host` is the most specific part of the connection so probably identifies it best even if not technically correct. You can leave it as is if you want.

samuong: Looking at this a bit closer, I feel like these error messages shouldn't get sent back in a 500 response body anyway - they should be logged together with the request id. We probably don't even need to send a response body back at all, so this could just change to `w.WriteHeader(http.StatusInternalServerError)`.

There are a number of places that this happens elsewhere in the code, so I'll go through and make the changes in a separate pull request.

PR #24, opened 2019-09-18
Use assert/require.NoError() instead of Nil()

samuong:


samuong:
I never really knew that there was a `NoError()` function until recently, so I was using `Nil()` instead. This pull request changes all the calls to `Nil(t, err)` to `NoError(t, err)`

samuong:
Ah, good catch. I was just doing a blind search and replace for `Nil(t, err)` and had forgotten the cases where I wasn't checking an actual variable named `err`. This is fixed now. The remaining `assert.Nil()` calls are for non-`error` values, so I've left them.

PR #28, opened 2019-09-22
Log errors instead of sending them back in the response body

samuong:


proxy.go
```diff
@@ -158,7 +167,8 @@ func (ph ProxyHandler) proxyRequest(w http.ResponseWriter, req *http.Request, au
 			req.Body = ioutil.NopCloser(rd)
 			resp, err = auth.do(req, ph.transport)
 			if err != nil {
-				http.Error(w, err.Error(), http.StatusInternalServerError)
+				log.Printf("[%d] Error forwarding request (with auth): %v", id, err)
```
camh-: Unfortunately github does not allow me to comment on lines outside the diff context.

5 lines above is an error being logged, but it does not have the id, and it also does not write an error or return. Not sure if being unable to seek is a terminal error, but perhaps at least log the id now that you have it.

proxy.go
```diff
@@ -158,7 +167,8 @@ func (ph ProxyHandler) proxyRequest(w http.ResponseWriter, req *http.Request, au
...
+				w.WriteHeader(http.StatusInternalServerError)
```
camh-: Unfortunately github does not allow me to comment on lines outside the diff context.

13 lines below this you are logging an error with the id, but you're pulling it out of the context. You can use `id` there now since you have it.

Also in other places you converted logging errors with `%v` - that error log still has `%s`.

proxy.go
```diff
@@ -51,9 +51,12 @@ func (ph ProxyHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 }
 
 func (ph ProxyHandler) handleConnect(w http.ResponseWriter, req *http.Request) {
+	// Establish a connection to the server, or an upstream proxy.
 	u, err := ph.transport.Proxy(req)
+	id := req.Context().Value("id")
 	if err != nil {
-		http.Error(w, err.Error(), http.StatusInternalServerError)
+		log.Printf("[%d] Can't find proxy for %v: %v", id, req.Host, err)
```
camh-: Most of the other errors messages start with "Error". Should you standardise on that for errors being logged? If so this could be "Error finding proxy for..."

proxy.go
```diff
@@ -66,23 +69,27 @@ func (ph ProxyHandler) handleConnect(w http.ResponseWriter, req *http.Request) {
 		w.WriteHeader(http.StatusBadGateway)
 		return
 	}
+	// Establish a connection back to the client, by hijacking the ResponseWriter.
 	h, ok := w.(http.Hijacker)
 	if !ok {
-		msg := fmt.Sprintf("Can't hijack connection to %v", req.Host)
-		http.Error(w, msg, http.StatusInternalServerError)
+		log.Printf("[%d] Can't hijack response writer", id)
```
camh-: Likewise, this could be "Error converting response writer into a hijacker"

proxy.go
samuong: Done

proxy.go
samuong: Done

proxy.go
samuong: Done, thanks for catching that

proxy.go
samuong: Done and done

samuong:
I've made all the changes you mentioned, can you take another look?

Issue #29, opened 2019-09-24
The Google Cloud SDK (gcloud) doesn't work through Alpaca


samuong:
Steps to reproduce:

1. Set up the proxy to go via Alpaca, by following the instructions at: https://cloud.google.com/sdk/docs/proxy-settings#proxy_configuration
2. Run `gcloud compute addresses list`. If you haven't already done so, it'll ask you to provide your credentials and project name.

The output should look like this:

```
$ gcloud compute addresses list
Listed 0 items.
```

But what actually happens is that it just sits there doing nothing.

samuong:
This regression started with 42e94be.

samuong:
Turns out this is due to the wrong newline character being used. It should be `\r\n`, not just `\n`, according to https://www.w3.org/Protocols/rfc2616/rfc2616-sec2.html#sec2.2.

PR #30, opened 2019-09-24
Construct and write an http.Response rather than a hardcoded string

samuong:
Fixes: #29

### DIFF
--- a/proxy_test.go
+++ b/proxy_test.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bufio"
 	"crypto/tls"
+	"strings"
 	"crypto/x509"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


proxy_test.go
```diff
@@ -289,3 +290,31 @@ func testConnectResponseHeaders(t *testing.T, server string, client net.Conn) {
 	assert.Empty(t, resp.TransferEncoding)
 	assert.Equal(t, int64(-1), resp.ContentLength)
 }
+
+func TestConnectResponseHasCorrectNewlines(t *testing.T) {
```
camh-: This test doesn't really belong in this code base because it is not responsible now for the correct formatting of responses. But obviously you want the test because it had the bug. So perhaps a comment on why this test is here might be warranted. Up to you.

samuong: Yeah I was on the fence about adding a test, and thought it might be good to have one in case anyone decides that building our own `http.Response` and writing it out is overkill compared to just writing a hard-coded string. I've changed the comment slightly, hope it's clear enough...

camh-: I actually completely missed the link to the issue. I just saw the rfc link and thought it was all the same. Better now

PR #36, opened 2019-11-24
Return a pointer to an authenticator

samuong:
This allows Alpaca to distinguish between having no credentials, and having a blank domain/username/password.


samuong:
Makes sense, done. I've also rebased this onto the other change which modified the logs slightly (to remove the newline characters).

Issue #8, opened 2019-06-01
During network change, alpaca incorrectly determines that proxy is unreachable


samuong:
Alpaca monitors the host's IP addresses to determine whether to use the PAC file, or proxy requests directly. For example, if a user moves from a corporate network to their home WiFi, their IP addresses will change, Alpaca will notice that the PAC server is no longer reachable, and conclude that requests should be sent directly.

This heuristic can fail when a new IP address is issued, but DNS resolution fails for some time until the network connection is fully established (which could happen in cases of slow 802.1x negotiation?)

Chrome works around this by [waiting 2 seconds](https://cs.chromium.org/chromium/src/net/proxy_resolution/proxy_resolution_service.cc?l=96&rcl=3db5f65968c3ecab3932c1ff7367ad28834f9502) after the IP address change before attempting to fetch the PAC file. However Chrome seems to have a way to get notified of IP address changes, whereas Alpaca checks the IP addresses when it does a proxy request.

So another option might be to introduce a retry mechanism when downloading the PAC file, with an exponential backoff.

samuong:
After a bit more research, Chrome reconfigures its proxy settings 2 seconds after it detects that either the addresses have changed, or the DNS configuration has changed (see https://cs.chromium.org/chromium/src/net/proxy_resolution/proxy_resolution_service.h - `ProxyResolutionService` extends both `NetworkChangeNotifier::IPAddressObserver` and `NetworkChangeNotifier::DNSObserver`).

Go doesn't seem to provide a portable way to monitor DNS config changes, at least not through the standard library (although third-party libraries like github.com/miekg/dns do exist) so it would be hard to replicate what Chrome does.

PR #37, opened 2019-12-01
Retry PAC downloads after 2 seconds in case of failure

samuong:
I can't think of a better way to fix issue #8. This should work around it in most, if not all, cases.

### DIFF
--- a/pacfetcher_test.go
+++ b/pacfetcher_test.go
@@ -18,6 +18,11 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func init() {
+	// Set the retry delay to zero, so that it doesn't delay unit tests.
+	delayAfterFailedDownload = 0
+}
+
 func pacjsHandler(pacjs string) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) { w.Write([]byte(pacjs)) }
 }
--- a/pacfetcher_test.go
+++ b/pacfetcher_test.go
@@ -23,12 +23,12 @@ func pacjsHandler(pacjs string) http.HandlerFunc {
 }
 
 type pacServerWhichFailsOnFirstTry struct {
-	firstTry bool
+	count int
 }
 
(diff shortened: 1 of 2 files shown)


samuong:
I agree. But years of being trained by tools like SonarQube (and similarly-programmed humans) has made me feel guilty every time I don't create a named constant. I've just changed this so that the delay is inline with the code, which I think is easier to read, and doesn't result in a false sense of security about the consistency of the log message and the actual delay.

I've also added some more error checking. Previously the call to `log.Printf()` could have been called with a nil error value. It was also possible that a retried download would leak a `resp.Body`.

PTAL?

pacfetcher_test.go
```diff
@@ -22,6 +22,19 @@ func pacjsHandler(pacjs string) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) { w.Write([]byte(pacjs)) }
 }
 
+type pacServerWhichFailsOnFirstTry struct {
```
camh-: If this type were `pacServerWhichFailsFirstNTries` and was initialised with a count, and decremented that count erroring until the count was 0, you could also use it to test the untest case of 2+ failures. But that wasn't tested before so no real need to test it now. Up to you.

samuong: I think the interesting equivalence classes are: 1) consistently failing downloads, 2) consistently succeeding downloads, and 3) failing downloads that at some point start to work. Having 1, 2 or n initial failures before the first successful download all seem to be the same as the third class of tests to me.

Also, I'm a little hesitant to do this because the timeout is 2 seconds, not a fixed number of failures, and if we go to a high-enough number of fixed failures then we'll end up with a flaky test.

samuong:
While thinking about your last comment, I realised that this PR adds about 6 seconds to the time taken to run the tests (it goes from less than half a second to over six seconds to run `go test`). So I'm going to hold off on merging this, and I'm planning to make the 2 second sleep a parameter inside the `pacFetcher` struct, and setting it to something else for unit tests.

samuong:
@camh- I've fixed the tests, can you please take another look?

PR #38, opened 2019-12-17
Add Apache 2.0 license notice to *.go, and add an AUTHORS file

samuong:


samuong:
To be honest, I didn't put a huge amount of thought into the different licenses. When I got my first pull request for Alpaca (#7), I had to pick a license so that we had _something_. Apache 2.0 seemed like a good default, but I didn't do an exhaustive comparison of all of the possible licenses.

The appendix in the LICENSE file seems to say that this boilerplate "text should be enclosed in the appropriate comment syntax for the file format". But IANAL so let me know if you think I've interpreted this incorrectly.

juliaogris:
wow, that initialism will put a smile on face before i go to bed tonight. which is now.
i think you have done the right thing. would be reason for me not to use Apache 2.0 though, don't like these repetitive headers.
merge-on.

samuong:
Thanks Julia. It looks like the Apache Foundation itself has a policy of putting this notice on every source file for its own projects: http://www.apache.org/legal/release-policy.html#which-files-contain-license

But Alpaca is not an Apache Foundation project - it's just licensed under an Apache License. I'm not 100% sure if a LICENSE file in the root (and no boilerplate on each file) implies that the license covers the entire repo. But it seems like it's better to be clear about it than not.

PR #42, opened 2019-12-19
Avoid calling WriteHeader on a hijacked connection

### LINKED ISSUE #39: WriteHeader is called on a hijacked connection
During the handling of a CONNECT request, if the response fails to be written, we will [call net/http.ResponseWriter.WriteHeader()](https://github.com/samuong/alpaca/blob/0d69c42f875313e7e3f76054a8bc4939ac3562f1/proxy.go#L103) on an already-hijacked connection. The call should just be removed (the error is already being logged).

samuong:
Fixes #39


samuong:
I thought putting "Fixes #39" would be enough - does the issue explain the reason why in enough detail? I'm hesitant to duplicate this in the git log message, but let me know if you think there's a good reason to do this?

Issue #20, opened 2019-08-28
Setup CI


juliaogris:
- [ ] setup CI for all PRs: 
  - build
  - test
  - lint
  - coverage (we don't want it to drop)
- [ ] setup required checks to pass in `master` branch protection
- [ ] create CONTRIBUTING.md, describing 
- [ ] create Makefile so that `make` resembles CI run as closely as possible

PR #45, opened 2020-01-19
Github actions cleanup and native build. 

jamesmoriarty:
- CI on push.
- Release on tag push prefixed with v*.
- Native build release binaries.

Screen Shot 2020-01-20 at 8 34 12 pm

Screen Shot 2020-01-19 at 6 57 41 pm


.github/bin/tag
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+[ "$#" -ge "1" ] || { echo "usage: $0 v1.0.11"; exit 1; }
+
+set -x
+
+tag=$1
+
+git tag --delete $tag
+git push origin :$tag
```
samuong: I'm not sure that blindly deleting an existing tag is a good idea; wouldn't it be better to error out in the case that a tag already exists? If you delete and recreate it, it'll force a new release, which would overwrite existing release binaries, which might've already been downloaded by users.

More broadly though, I'm not sure this script adds a lot of value. Creating and pushing a tag is easy enough to do manually, and those looking at a simpler way could just use the GitHub UI.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,24 @@
...
+        go: [ '1.13' ]
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-go@v1
+        with:
+          go-version: ${{ matrix.go }}
+
+      - run: gofmt -l $(find . -type f -name '*.go' -not -path "./vendor/*") 2>&1
```
samuong: Do we have access to `goimports` in GitHub Actions? If so, it'd be better to use that. A while ago we made the decision to use that for formatting, since it separately groups the official and third-party import packages, and removes unused imports (otherwise it's the same as `gofmt`).

@camh- and/or @juliaogris might also want to use `golangci-lint`, if it's available?

Also, there's no `vendor` directory in the Alpaca codebase (we're using Go Modules) so you don't need the `-not -path "./vendor/*"`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,24 @@
...
+
+      - run: go build -v .
+
+      - run: go test .
```
samuong: There are some tests inside the `cancelable` subpackage that won't run if you do it like this, so you should use `go test ./...` instead.

pacrunner_test.go
```diff
@@ -313,6 +313,7 @@ func TestWeekdayRange(t *testing.T) {
 	for _, test := range tests {
 		for i, weekday := range weekdays {
 			t.Run(test.name+" "+weekday.name, func(t *testing.T) {
+				t.Skip("Pending test to implement Github Releases.")
```
samuong: What was the reason for skipping these tests?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,24 @@
+name: Continuous Integration
+
+on: [ push ]
+
+jobs:
+  test:
+    runs-on: ubuntu-16.04
```
samuong: Is there a way to separately run this on macos-latest as well? I've accidentally broken the build on macOS before (see #40) because I develop on Linux and it wasn't running some darwin-only tests.

.github/workflows/release.yml
```diff
@@ -0,0 +1,72 @@
...
+
+  build:
+    runs-on: macos-latest
+
+    needs: [ release ]
+
+    strategy:
+      matrix:
+        go: [ '1.13' ]
+        goos: [ 'linux', 'darwin' ]
```
samuong: Does this mean that all of the release builds happen on macOS, and the Linux build gets cross-compiled? I would be worried that this won't work with cgo - in particular, the go-keychain package uses cgo, and it's not possible to cross-compile a macOS binary from a Linux builder. It might be possible to produce a Linux binary from macOS though...

If it's difficult to do this, and you can confirm that it produces a working Linux binary, then we can probably leave it and deal with it later (i.e. when we introduce a cgo package that doesn't build on macOS).

samuong:
Also it'd be good to somehow link this to #20. Not sure if this comment itself is enough to do it, or if it needs to be in the git log comment...

.github/workflows/ci.yml
jamesmoriarty: Changed to:

```
jobs:
  test:
    strategy:
      matrix:
        os: [ 'macos-latest', 'ubuntu-16.04' ]
        go: [ '1.13' ]

    runs-on: ${{ matrix.os }}
```

.github/workflows/release.yml
jamesmoriarty: Thanks for this. I've changed it to build on the native os. There was an issue with crossbuild with:

```
        env:
          GOOS: ${{ matrix.target.goos }}
          GOARCH: ${{ matrix.target.goarch }}
          CGO_ENABLED: 1
```

I've added:

```
    strategy:
      matrix:
        target:
          - os: 'macos-latest'
            goos: 'darwin'
            goarch: 'amd64'
          - os: 'ubuntu-16.04'
            goos: 'linux'
            goarch: 'amd64'
        go: [ '1.13' ]
  
    runs-on: ${{ matrix.target.os }}
```

and it looks good.

.github/bin/tag
jamesmoriarty: Removed. Useful for testing releases.

pacrunner_test.go
jamesmoriarty:
```
            	Error:      	Not equal: 
            	            	expected: false
            	            	actual  : true
            	Test:       	TestWeekdayRange/M-F_UTC_MON
    --- FAIL: TestWeekdayRange/M-F_UTC_SAT (0.00s)
        pacrunner_test.go:327: 
            	Error Trace:	pacrunner_test.go:327
            	Error:      	Not equal: 
            	            	expected: true
            	            	actual  : false
            	Test:       	TestWeekdayRange/M-F_UTC_SAT
    --- FAIL: TestWeekdayRange/SAT_UTC_SUN (0.00s)
        pacrunner_test.go:327: 
            	Error Trace:	pacrunner_test.go:327
            	Error:      	Not equal: 
            	            	expected: true
            	            	actual  : false
            	Test:       	TestWeekdayRange/SAT_UTC_SUN
    --- FAIL: TestWeekdayRange/SAT_UTC_SAT (0.00s)
        pacrunner_test.go:327: 
            	Error Trace:	pacrunner_test.go:327
            	Error:      	Not equal: 
            	            	expected: false
            	            	actual  : true
            	Test:       	TestWeekdayRange/SAT_UTC_SAT
FAIL
FAIL	github.com/samuong/alpaca	0.271s
ok  	github.com/samuong/alpaca/cancelable	0.003s
```

Will have to investigate more but this is what I'm seeing.

samuong: That's really weird. I've created #46 to track this, but the problem isn't with your change, so let's disable it in this pull request for now.

I suspect this has something to do with timezones not being handled properly - I'm pretty sure it worked when I ran it in Melbourne, but now it fails on my local machine in New York.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,66 @@
...
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-go@v1
+        with:
+          go-version: ${{ matrix.go }}
+ 
+      - run: gofmt 2>&1
```
samuong: I think you've removed too much? If you just have `gofmt 2>&1` won't it just hang there waiting until it finds something on stdin? I'd expect this to look something like:

```
gofmt -l $(find . -type f -name '*.go') 2>&1
```

Also, I guess `goimports` is not available? If not, we can use `gofmt` for now.

AUTHORS
```diff
@@ -1,3 +1,4 @@
+James Moriarty <jamespaulmoriarty@gmail.com>
```
samuong: Nit: this file should have entries that are sorted alphabetically

.github/workflows/ci.yml
samuong: I just took a look at the run in your repo, and it looks like it didn't hang waiting for input, but it did finish in 0s (and didn't seem to do anything). This might be because GitHub connects an empty input stream to the job?

https://github.com/jamesmoriarty/alpaca/runs/398642375

One way to test this would be to push a commit with some bad formatting, and see if it complains.

pacrunner_test.go
jamesmoriarty: Added ref to issue in pending.

AUTHORS
jamesmoriarty: 👍

.github/workflows/ci.yml
jamesmoriarty: Thanks for catching this. Updated e.g. https://github.com/jamesmoriarty/alpaca/runs/400039825

jamesmoriarty:
Thanks for the feedback @samuong. I appreciate it.

jamesmoriarty:
Screen Shot 2020-01-21 at 1 32 27 pm

Hey @samuong, I can't merge but am happy for you too.

samuong:
Oh right, ok I'll do it for you.

Issue #32, opened 2019-10-19
Alpaca fails to get credentails from NoMAD/Keychain


samuong:
Even when NoMAD is installed and configured to use the system keychain, Alpaca fails to read the NoMAD configuration and get the credentials out of the keychain. It disables proxy authentication as a result:

```
main.go:58: NoMAD configuration key not found. Disabling proxy authentication.
```

When NoMAD app preferences are managed by a configuration profile which means they get written to `/Library/Managed Preferences/$USER/com.trusourcelabs.NoMAD.plist`. When a user then logs into NoMAD, their particular details get written to `/Users/$USER/Library/Preferences/com.trusourcelabs.NoMAD.plist`. In this case, `UseKeychain` gets defined in `Managed Preferences` and `UserPrincipal` gets defined in the user's `Preferences` directory.

Alpaca uses the `defaults` command to read the configuration, and it looks like this is only reading the plist file in the user's `Preferences` directory.

A workaround is to manually use `defaults write` to set `UseKeychain` to `1` in the user's plist:

```
defaults write com.trusourcelabs.NoMAD UseKeychain 1
```

PR #50, opened 2020-03-06
Read NoMAD configuration values from Managed Preferences

segan5:
Read NoMAD configuration values from /Library/Managed Preferences before reading them from ~/Library/Preferences. 

I've done some testing and it all seems to work, `UseKeychain` is now being read from Managed Preferences plist while `UserPrincipal` is still being read from the user's preferences plist.

This will avoid users having to use that manual workaround on corporate managed devices that deploy a configuration profile to manage NoMAD.

Fixes: #32


nomad_darwin.go
```diff
@@ -31,12 +31,19 @@ func init() {
 }
 
 func readDefaultForNoMAD(key string) (string, error) {
-	cmd := execCommand("defaults", "read", "com.trusourcelabs.NoMAD", key)
-	out, err := cmd.Output()
-	if err != nil {
-		return "", err
+	// check for key in managed preferences first
+	cmdManaged := execCommand("defaults", "read", "/Library/Managed Preferences/com.trusourcelabs.NoMAD.plist", key)
+	outManaged, errManaged := cmdManaged.Output()
```
camh-anz: generally you should always call your error variable `err` unless you actually need two errors to co-exist.

I would also still just use `cmd` and `out` as you can reuse them, or since you're using `:=` on the inner scope, redefine them.

nomad_darwin.go
```diff
@@ -31,12 +31,19 @@ func init() {
...
+	if errManaged != nil {
+		// now check user preferences
+		cmdUser := execCommand("defaults", "read", "com.trusourcelabs.NoMAD", key)
+		outUser, errUser := cmdUser.Output()
+		if errUser != nil {
+			return "", errUser
+		}
+		return strings.TrimSpace(string(outUser)), nil
```
camh-anz: Ideally the "happy" return is at the end of the function (as per https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88 - "Put the happy return statement as the very last line"). You could make this happen by using the same name for `out` and not redefining it on the inner loop: `out, err := cmdUser.Output()` - then you could just fall out of the `if` to the return at the end and it will return the right thing.

nomad_darwin.go
```diff
@@ -31,12 +31,19 @@ func init() {
...
+		// now check user preferences
+		cmdUser := execCommand("defaults", "read", "com.trusourcelabs.NoMAD", key)
+		outUser, errUser := cmdUser.Output()
+		if errUser != nil {
+			return "", errUser
+		}
+		return strings.TrimSpace(string(outUser)), nil		
 	}
-	return strings.TrimSpace(string(out)), nil
+	return strings.TrimSpace(string(outManaged)), nil
```
camh-anz: Combining my previous suggestions with a little bit of repetition removal:
```
func readDefaultForNoMAD(key string) (string, error) {
        userDomain := "com.trusourcelabs.NoMAD"
        mpDomain := fmt.Sprintf("/Library/Managed Preferences/%s.plist", userDomain)

        // Read from managed preferences first
        out, err := execCommand("defaults", "read", mpDomain, key).Output()
        if err != nil {
                // Read from user preferences if not in managed preferences
                out, err = execCommand("defaults", "read", userDomain, key).Output()
        }
        if err != nil {
                return "", err
        }
        return strings.TrimSpace(string(out)), nil
}
```

samuong:
Thanks Ern! Don't forget to add yourself to `AUTHORS`. You should probably also change "2019" to "2019, 2020" at the top of `nomad_darwin.go` as well.

segan5:
Thanks for the comments @camh-anz and @samuong! Bit of a novice at go so I welcome all the feedback.

nomad_darwin.go
```diff
@@ -31,10 +32,17 @@ func init() {
...
+
+	// Read from managed preferences first
+	out, err := execCommand("defaults", "read", mpDomain, key).Output()
+	if err != nil {
+			// Read from user preferences if not in managed preferences
+			out, err = execCommand("defaults", "read", userDomain, key).Output()
+	}
 	if err != nil {
-		return "", err
+			return "", err
```
camh-: This and some previous lines are indented too far. This means you have not run `go fmt` on the file. If you want to do more Go development I recommend setting up your editor to run it every time you save. Even better is to install `goimports` (https://godoc.org/golang.org/x/tools/cmd/goimports) which does all that `go fmt` does but also adds and removes required imports. But if this is just a one-off Go exercise for you, just run `go fmt` over the file (`go help fmt` for more details).

AUTHORS
```diff
@@ -4,3 +4,4 @@ James Moriarty <jamespaulmoriarty@gmail.com>
 Julia Ogris <julia.ogris@gmail.com>
 Keith Ferguson <keith.ferguson@anz.com>
 Sam Uong <samuong@gmail.com>
+Seng Ern Gan <sengern.gan@anz.com>
```
camh-: You're missing an end of line at end of file (see the red marker in the github review page). Please add. https://gist.github.com/camh-/1bebfcff1b0f814e9b191edc60d5206b

segan5:
Thanks @camh- actually I don't think I can merge the PR, it is saying I don't have write access to the repo. Which is fine by me. So if someone with write access can merge it when convenient, that would be great.

PR #51, opened 2020-03-25
Add releases to readme

jamesmoriarty:
Screen Shot 2020-03-26 at 4 00 48 pm

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Alpaca
 
+![Latest Tag][2] ![GitHub Workflow Status][3] ![GitHub Releases][4]
+
 Alpaca is an HTTP/HTTPS proxy. It is designed for users of Unix tools, who
 operate inside networks that use proxy auto-configuration (PAC) files and
 require NTLM authentication.
@@ -8,15 +10,19 @@ It currently does not yet implement many important features, such as SOCKS,
 HTTP/2, and many of the predefined JS functions that PAC files use.
 Contributions are welcome, please reach out to me if you'd like to help!
 
-## Installation and Usage
+## Download Binary
 
-To download and install Alpaca, use:
+Alpaca can be downloaded from the [GitHub releases page][1].
+
+## Build from Sources
 
 ```sh
 $ go get -v -u github.com/samuong/alpaca
 ```
 
-Then start Alpaca by running the `alpaca` binary.
+## Usage
+
+Start Alpaca by running the `alpaca` binary.
 
 If your proxy requires NTLM authentication, you'll need to supply your domain and
 username (via command-line flags) and a password (via a prompt):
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,14 +10,18 @@ It currently does not yet implement many important features, such as SOCKS,
 HTTP/2, and many of the predefined JS functions that PAC files use.
 Contributions are welcome, please reach out to me if you'd like to help!
 
-## Installation and Usage
+## Download
+
+Releases can be downloaded from [here][1].
```
camh-: Using the string "here" as an anchor for a link is poor. Better to describe what is "here":

Releases can be downloaded from the [alpaca github releases page][1].

That way the link text describes the link, which can also be useful for accessibility.

README.md
```diff
@@ -8,14 +10,18 @@ It currently does not yet implement many important features, such as SOCKS,
...
 
-To download and install Alpaca, use:
+## Install
```
camh-: Perhaps "Install from source"?

README.md
```diff
@@ -8,14 +10,18 @@ It currently does not yet implement many important features, such as SOCKS,
...
 
 ```sh
 $ go get -v -u github.com/samuong/alpaca
 ```
 
+## Usage
+
 Then start Alpaca by running the `alpaca` binary.
```
camh-: Remove "Then" now that this is a new section with a title.

README.md
```diff
@@ -8,15 +10,19 @@ It currently does not yet implement many important features, such as SOCKS,
 HTTP/2, and many of the predefined JS functions that PAC files use.
 Contributions are welcome, please reach out to me if you'd like to help!
 
-## Installation and Usage
+## Download
+
+Alpaca can be downloaded from the [github releases page][1].
 
-To download and install Alpaca, use:
+## Install From Source
```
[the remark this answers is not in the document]
camh-: I'm in two minds about this. I like to know that a go program can be installed with `go get` - too many cant - they need a makefile or some generate step. Also, people are looking at the README from the source code - i think the programmer is the primary audience of that. Other documentation can be more user-oriented.

However I think it would help make it less ambiguous if the "Download" section was called "Install from GitHub Release" or something, so the "Install from Source" stands out as an alternate installation option. Having a Download then Install section will lead the less discerning user to try one after the other I would expect.

Perhaps the "Install from GitHub Release" should also say a bit more than download the release. Unpack and put on your path is also needed, in whatever detail that would be required for the audience.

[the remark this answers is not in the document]
samuong: OK, what if we called the sections "Download Binary" and "Build from Sources"? Hopefully that should be obvious enough.

I agree that it's not always obvious that a Go program is `go get`-able, but given the way GitHub displays the README file on the project page, this is the closest thing to a "home page" that Alpaca has, which is why I think of this as a user-oriented file.

PR #56, opened 2020-06-30
Add instructions to install via Homebrew

### LINKED ISSUE #55: Provide a Brew tap
It would be super useful for alpaca to have a brew tap so it can be easily installed and updated without go installed.

samuong:
Also, I've moved the "Download Binary" method down one section, because I think
that Homebrew and Go both provide a better (automated) way to keep the binary
up-to-date.

Also, I've added `http://` to `http_proxy` and `https_proxy`. This seems to be
needed by `yarn`.

Fixes #55

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,15 +5,30 @@
 Alpaca is a local HTTP proxy for command-line tools. It supports proxy
 auto-configuration (PAC) files and NTLM authentication.
 
-## Download Binary
+## Install using Homebrew
 
-Alpaca can be downloaded from the [GitHub releases page][1].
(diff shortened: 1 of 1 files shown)


samuong:
Oh no, I accidentally copied-and-pasted an extra space at the start of the git commit message. I'm not sure if I can live with this; is there a way to fix it?

camh-:
git commit --amend

Your can reset the merge commit away and re merge

samuong:
There's no merge commit - I squashed and merged it when closed the PR. I'm guessing it's not possible to force push over a protected branch, right?

PR #57, opened 2020-09-28
Add godocs for the cancelable package, and remove an unnecessary call to server.Close()

samuong:
This PR has two separate commits, and will be merged using a fast-forward merge.

### DIFF
--- a/cancelable/closer.go
+++ b/cancelable/closer.go
@@ -1,4 +1,4 @@
-// Copyright 2019 The Alpaca Authors
+// Copyright 2019, 2020 The Alpaca Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -12,24 +12,34 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package cancelable provides a cancelable version of io.Closer.
+//
+// This is intended to be used by functions where the responsibility for closing
+// a resource sometimes belongs to the function itself (e.g. early returns due
+// to error handling), and sometimes belongs to the caller.
 package cancelable
 
 import (
 	"io"
 )
 
+// Closer is an io.Closer that can be cancelled.
 type Closer struct {
 	c io.Closer
 }
 
+// NewCloser returns a cancelable.Closer that wraps an io.Closer.
 func NewCloser(c io.Closer) *Closer {
 	return &Closer{c: c}
 }
 
+// Cancel turns any future calls to Close into a no-op.
 func (c *Closer) Cancel() {
 	c.c = nil
 }
 
+// Close calls the Close function on the io.Closer (if it has not been cancelled
+// or already closed). It is safe to call this multiple times.
 func (c *Closer) Close() error {
 	if c.c == nil {
 		return nil
--- a/cancelable/example_closer_test.go
+++ b/cancelable/example_closer_test.go
@@ -1,0 +1,58 @@
+// Copyright 2020 The Alpaca Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cancelable_test
+
+import (
+	"fmt"
+
+	"github.com/samuong/alpaca/cancelable"
+)
+
+type printCloser struct {
+	msg string
+}
+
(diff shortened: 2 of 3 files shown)


cancelable/closer.go
```diff
@@ -12,24 +12,65 @@
...
 func NewCloser(c io.Closer) *Closer {
 	return &Closer{c: c}
 }
 
+// Cancel turns any future calls to Close into a no-op.
 func (c *Closer) Cancel() {
 	c.c = nil
 }
 
+// Close calls the Close function on the io.Closer. It is safe to call this
```
camh-, line 72: Not necessarily, only if it has not been cancelled or already closed.

cancelable/closer.go
```diff
@@ -12,24 +12,65 @@
...
+//
+// This is intended to be used by functions where the responsibility for closing
+// a resource sometimes belongs to the function itself (e.g. early returns due
+// to error handling), and sometimes belongs to the caller.
+//
+// In the example below, this responsibility is shared by both f() and caller().
+// There is always a deferred function that is ready to close the connection
+// when needed, and we never need to worry about forgetting to call Close().
+//
+//  func f() (net.Conn, error) {
```
camh-, line 25: You could put this in `example_closer_test.go` so it can function as a test and show as an example in the docs. I thought of this as I started reviewing the code in the comment, wondering if there were any typos. Although I admit a real example might be hard in the playground environment that examples run in.

If you haven't looked at testable examples before, start here: https://blog.golang.org/examples

You can see what we have done at https://pkg.go.dev/foxygo.at/s/httpe (example source at https://github.com/foxygoat/s/tree/master/httpe)

cancelable/closer.go
samuong: Done

cancelable/closer.go
samuong: Yeah I was thinking of doing an example, but then I realized that the "play" button wouldn't work due to the sandbox not allowing any network requests. Happy to put up an example if you can think of another Closer that could be wrapped, and works in the sandbox...

camh-: You could just knock up an example closer that prints "Closed" (or not, if cancelled). It would only be a couple of extra lines.

camh- (dismissed):
LGTM. I'll leave whether to do an example or not up to you.

cancelable/closer.go
samuong: Makes sense, I've just done that. But it doesn't seem to work when I add `//output:` comments, or when I try to have multiple examples (possibly because I've got a too-old version of `godoc`?). Hopefully the example isn't too noisy, let me know what you think?

samuong:
@camh- not sure what went wrong earlier with the checks, but I think it was due to me force pushing (to change the example's package name from `cancelable` to `cancelable_test`) while the checks were running, which confused GitHub Actions.

Can you take another look?

camh-:
Haven't had much of a chance to look yet, but my first thoughts were that the example was too complicated. I know you kind of want to make it look real, so you have `frob()` which conditionally returns an error, but I think that obscures what is going on. I think you just need to demonstrate that a `defer close` sometimes calls close and sometimes does not. I also think it is important to get the output stuff working, otherwise nothing is tested and it may not even work.

I'll have a quick play and see.

camh-:
Not sure if you think this is any better, but it is what I had in mind:
```
import (
        "fmt"

        "github.com/samuong/alpaca/cancelable"
)

type printCloser struct {
        msg string
}

// Close prints the printCloser's msg on stdout if it has one and clears it
// so it doesn't print it again.
func (pc *printCloser) Close() error {
        if pc.msg != "" {
                fmt.Println(pc.msg)
                pc.msg = ""
        }
        return nil
}

func ExampleCloser() {
        c := create(false)
        c.Close() // prints nothing - already closed in create()

        c = create(true)
        c.Close() // prints msg - not closed in create()

        // Output:
        // deferred close not cancelled
        // deferred close cancelled
}

func create(doCancel bool) *printCloser {
        pc := &printCloser{"deferred close not cancelled"}

        closer := cancelable.NewCloser(pc)
        defer closer.Close()

        if !doCancel {
                return pc // deferred close will cause msg to be printed
        }

        closer.Cancel()
        pc.msg = "deferred close cancelled"
        return pc
}
```

I liked your `runtime.Caller()` thing, but I couldn't make it work.

samuong:
I like it, thanks! And it passes the test.

And yeah, I wrote all the scaffolding like the `frob()` stuff before I realized that it wouldn't be hidden from the example. I think it looks a lot better the way you suggested.

camh-:
I realised what it should do as I was drifting off to sleep. The `printCloser` should not have that empty string test or print only once. Just print the message. That way, we can see the first one being printed twice and the second printed only once. I would initialise the string with `fmt.Sprintf("doCancel: %v", doCancel)` so you would see `doCancel: false` printed twice, but `doCancel: true` printed once.

samuong:
@camh- sorry for the delay, I've taken your example from Slack and left it as a single example function. Not sure I really want the two examples - which are very similar - to be split further apart on the doc page. But we can revisit this later if anyone thinks it's not clear enough.

camh- (approved):

PR #59, opened 2020-10-15
Remove bare string used as context key

### LINKED ISSUE #17: Remove bare string used as context key.
The following code uses a bare string as a context key. This is a [bug](https://golang.org/pkg/context/#WithValue).

https://github.com/samuong/alpaca/blob/8ac2ded38992c1795572fe4aba38eafaed7521c7/proxyfinder_test.go#L35

https://github.com/samuong/alpaca/blob/7d23c7f4fa7937c229d0cf5fe65a77d0dfa54abb/proxy.go#L41

samuong:
Also, use sync/atomic's AddUint64() to generate IDs, which is simpler
and avoids tying up a goroutine.

Fixes #17

### DIFF
--- a/contextid.go
+++ b/contextid.go
@@ -17,22 +17,22 @@ package main
 import (
 	"context"
 	"net/http"
+	"sync/atomic"
 )
 
-// AddContextID wraps a http.Handler to add a strictly increasing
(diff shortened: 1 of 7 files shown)


camh- (approved):
LGTM

PR #60, opened 2021-04-30
Modernisation

VeryStrongFingers:
Just some housekeeping & binaries for Windows & M1 MacBooks
(unfortunately GitHub macos-latest runners are still at OSX 10.5 so some additional steps had to be added)

- [GitHub action] release workflow now includes windows & darwin/arm64 (Apple M1 SoC)
- [GitHub action] CI workflow no longer triggers on tags being pushed
- [GitHub action] CI workflow to be triggered by pull requests
- [GitHub action] Updated Golang version 1.16 for builds (1.13 is EOL)
- [packages] `github.com/keybase/go-keychain` updated to fix deprecation notice below
> cgo-gcc-prolog:203:11: warning: 'SecTrustedApplicationCreateFromPath' is deprecated: first deprecated in macOS 10.15 - No longer supported [-Wdeprecated-declarations]
updated github.com/keybase/go-keychain to latest SHA (no tags yet :( )


.github/workflows/ci.yml
```diff
@@ -60,7 +72,18 @@ jobs:
...
-      - run: go build -v .
+      # xcode required until github macos runner is 11.x based
+      - uses: maxim-lobanov/setup-xcode@v1
+        if: matrix.target.goos == 'darwin' && matrix.target.goarch == 'arm64'
+        with:
+          xcode-version: latest
+
+      - if: matrix.target.goos == 'darwin' && matrix.target.goarch == 'arm64'
+        run: |
+          echo SDKROOT=$(xcrun --sdk macosx --show-sdk-path) >> $GITHUB_ENV
```
samuong, line 83: How come we don't need this for `amd64`? I'm guessing this is some quirk of how GitHub set up their macOS builders?

samuong (approved):
This is awesome, thanks!

.github/workflows/ci.yml
VeryStrongFingers: Pretty much, the default GitHub ubuntu/macos runners come with Xcode already installed but it's an older version before darwin/arm64

samuong: Fair enough, will just need to keep an eye out for when they update their builders. I want to land a few cleanups before the next release, and we'll be able to push something out (with Windows and M1 binaries) soon. Thanks again!

PR #61, opened 2021-05-03
Introduce a `transport` type to write HTTP requests to a `net.Conn`

samuong:
The purpose of this change is to reduce code duplication in
`authenticator.go`. Previously, we needed to have two separate
implementations of the NTLM handshake: one for CONNECT requests (which
are written directly to a `net.Conn`) and another for all other requests
(which are sent via an `http.Transport`). Now that we only need one
(written against any `http.RoundTripper`), it will be simpler to
implement additional authentication protocols in the future, such as
Kerberos.

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -1,0 +1,72 @@
+// Copyright 2021 The Alpaca Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


transport.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if t.conn == nil {
+		return nil, errors.New("no connection, can't send request")
+	}
+	if err := req.Write(t.conn); err != nil {
+		return nil, err
+	}
+	resp, err := http.ReadResponse(t.reader, req)
```
camh-, line 52: This could just be `return http.ReadResponse(t.reader, req)`. If you still prefer the error checking, change the final line to read `return resp, nil` (i.e. s/err/nil) since it is the non-error case and seeing the literal `nil` makes that easier to recognise.

camh- (approved):
just one little nitpick - feel free to ignore. i didn't realise that auth stuff was so thoroughly duplicated. I think my eyes glazed over when i looked at that...

transport.go
samuong: Done

camh- (approved):

PR #62, opened 2021-05-03
Keep a list of non-responsive proxies and avoid using them for 5 minutes

samuong:


proxy.go
```diff
@@ -239,3 +251,24 @@ func copyResponseHeaders(w http.ResponseWriter, resp *http.Response) {
...
+
+func (e *dialError) Error() string {
+	return fmt.Sprintf("error while dialing %q %q: %s", e.network, e.address, e.err.Error())
+}
+
+func (e *dialError) Unwrap() error {
+	return e.err
+}
+
+func dialContext(ignored context.Context, network, address string) (net.Conn, error) {
```
camh-, line 268: "ignored" is typically spelt "_" in Go

proxyfinder.go
```diff
@@ -85,34 +89,36 @@ func (pf *ProxyFinder) findProxyForRequest(req *http.Request) (*url.URL, error)
...
-	var host string
-	n, err := fmt.Sscanf(trimmed, "PROXY %s", &host)
-	if err == nil && n == 1 {
-		// The specified proxy should contain both a host and a port, but if for some reason
-		// it doesn't, assume port 80. This needs to be made explicit, as it eventually gets
-		// passed to net.Dial, which also requires a port.
-		proxy := &url.URL{Host: host}
-		if proxy.Port() == "" {
-			proxy.Host = net.JoinHostPort(host, "80")
+	arr := strings.Split(str, ";")
```
camh-, line 96: technically, strings.Split returns a slice not array, so this name is not quite right. you could just move it into the range since it is not used elsewhere that I can see:
```
for _, elem in range strings.Split(str, ";") {
```

proxyfinder.go
```diff
@@ -85,34 +89,36 @@ func (pf *ProxyFinder) findProxyForRequest(req *http.Request) (*url.URL, error)
...
+		} else if len(fields) == 2 && fields[0] == "PROXY" {
+			// The specified proxy should contain both a host and a port, but if for
+			// some reason it doesn't, assume port 80. This needs to be made explicit,
+			// as it eventually gets passed to net.Dial, which also requires a port.
+			proxy := &url.URL{Host: fields[1]}
+			if proxy.Port() == "" {
+				proxy.Host = net.JoinHostPort(proxy.Host, "80")
+			}
+			if pf.blocked.contains(proxy.Host) {
+				log.Printf("[%d] Skipping bad proxy: %q", id, proxy.Host)
```
camh-, line 111: this will produce a bit og log spam wont it? it will log on every request through alpaca when the first proxy on the list goes bad, until it comes good again?

I think instead you should just log when you put an entry on the block list and when you expire them.

proxyfinder.go
```diff
@@ -85,34 +89,36 @@ func (pf *ProxyFinder) findProxyForRequest(req *http.Request) (*url.URL, error)
...
+			log.Printf("[%d] %s %s via %q", id, req.Method, req.URL, elem)
+			return proxy, nil
 		}
-		return proxy, nil
-	}
-	if strings.HasPrefix(trimmed, "SOCKS ") {
-		return nil, errors.New("Alpaca does not yet support SOCKS proxies")
-	} else {
-		return nil, errors.New("Couldn't parse PAC response")
+		log.Printf("[%d] Coudln't parse proxy: %q", id, elem)
```
camh-, line 117: typo: Couldn't

Also, this will log for every request if the proxy string is malformed. it probably only needs to be logged once.

blocklist.go
```diff
@@ -0,0 +1,81 @@
...
+	entries []string             // Slice of entries, ordered by expiry time
+	expiry  map[string]time.Time // Map containing the expiry time for each entry
+	now     func() time.Time
+	mux     sync.Mutex
+}
+
+func newBlocklist() *blocklist {
+	return &blocklist{
+		entries: []string{},
+		expiry:  make(map[string]time.Time),
```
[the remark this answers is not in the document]
samuong: Done

proxy.go
samuong: Done

proxyfinder.go
samuong: Done

proxyfinder.go
samuong: That's a good point. I've logged this in both places before we block a proxy (rather than in `proxyFinder.blockProxy()`, because I also want to log the id).

proxyfinder.go
samuong: Fixed the typo. I think you're right that this is a bit spammy, but I might leave this here because it's a bit too much work to go and track which malformed proxy strings we've encountered.

Also, I think the most likely case that this would happen is an unsupported proxy scheme (e.g. SOCKS), in which case Alpaca is going to be unsuitable for the user. Alpaca should probably complain loudly in this case...

camh-: fair enough. perhaps you should log it in ALL CAPS as it is louder.

proxy.go
camh-: I started wondering why you didn't use the context parameter, and use `net.DialContext()` instead. So I looked where it was called from so I could understand what the context arg would be, and I could not find one. This function is unused.

BTW, golangci-lint also tells you this ;-)

camh-:
I wonder if this can or should be smarter? If all the proxies returned by the pac script are on the blocklist, alpaca will just do nothing for 5 minutes, for what may have been a brief blip in proxy reachability. Perhaps if all the proxies in a pac response are on the blocklist, ignore the blocklist?

proxy.go
samuong: Oops, looks like I forgot to clean that one up. We should add golangci-lint to GitHub Actions, and I've created #64 to track that.

samuong:
> I wonder if this can or should be smarter? If all the proxies returned by the pac script are on the blocklist, alpaca will just do nothing for 5 minutes, for what may have been a brief blip in proxy reachability. Perhaps if all the proxies in a pac response are on the blocklist, ignore the blocklist?

If all proxies are in the blocklist, we could:

1. Ignore (for this request) or clear (for all requests) the blocklist and use the first proxy from the PAC file
2. Not use a proxy and go direct
3. Clear the blocklist and try to download the PAC file again

The first option handles a transient network failure well, but might not be so good at handling the case where Alpaca is stuck thinking it's connected to the PAC server when it isn't (see issue #8). The second option is the other way around.

If we try to download the PAC file again (i.e. we basically re-evaluate whether we've got network connectivity) this helps with #8 as well.

I'm thinking of doing option 2 for this PR, and option 3 in a subsequent one. What do you think?

camh-:
I dont think going direct is going to help - if the pac file says to use a proxy, then i imagine it is unlikely to work without it.

I think if all proxies are on the block list, then you should just try the proxies as though they are not on the block list. If you are successful talking to a proxy, remove it from the block list.

A network change event should clear everything and the pac file gets refetched. I'm not sure I see the point of re-fetching the pac file when you have trouble talking to a proxy. If the network has not changed, then the pac file should still be valid.

samuong:
The specific case I'm talking about is where Alpaca incorrectly thinks it has connectivity to the PAC server (and the proxies that the PAC file references). In this case, it'll fail to dial any of the proxies, and eventually all of them will be added to the blocklist.

If it then falls back to DIRECT, then the request will make its way out to the public internet without trying to go via an unreachable proxy. If it falls back to the first proxy (ignoring the blocklist) then it'll be stuck forever. So in this case, wouldn't it be better to fall back to DIRECT?

But it could also be due to a transient network error as you mentioned. In that case, it's better to fall back to the proxy list, otherwise we'll be stuck trying to go DIRECT for the next 5 minutes.

Rather than try to guess which situation we're in, I think it would be better to check our connectivity status by trying to download the PAC file again.

camh-:
I dont understand this:

> The specific case I'm talking about is where Alpaca incorrectly thinks it has connectivity to the PAC server (and the proxies that the PAC file references)

If you know the proxies that the PAC file references, then you have connectivity to the PAC server, else how else do you have a PAC file to execute?

I assume you would not use a proxy to fetch the PAC file.

So, if you cannot reach the PAC server, you will be stuck with DIRECT, as you have no PAC file to tell you otherwise. If you get the PAC file, but cannot talk to the proxies, then you should not fall back to DIRECT because the PAC file does not tell you to go direct, so presumably that would just fail anyway.

Perhaps if all the proxies fail, you clear the block list, clear the PAC file and go back to the start and try to fetch the PAC file. If that fails, you will end up in DIRECT mode, but in the absence of a network change event, it suggests that connectivity to the PAC server will come back at some point when you should try to re-fetch.

I think we may have a different model of network failure in mind and maybe we are talking past each other. Since I don't use alpaca any more, I'm happy for you to try what you think works. You'll know if it turns out not to :)

proxy.go
samuong: Done (removed the unused `dialContext` function)

samuong:
Ok so as discussed, I think the cases that I'm thinking of are too much of an edge case, so I'm happy to ignore them for now. I went with your option of temporarily ignoring the blocklist in the event that all proxies are blocked.

camh- (approved):
LGTM

samuong:
During my own use of this branch, I haven't noticed any problems, so I'm going to land this now.

PR #63, opened 2021-05-03
Rename exported structs and functions

samuong:
This reduces the "public API" of the `main` package to nothing. The next
release of Alpaca will be v2.0.0, and any exported names will be made
available in sub-packages with a minor version bump.

### DIFF
--- a/authenticator.go
+++ b/authenticator.go
@@ -1,4 +1,4 @@
-// Copyright 2019 The Alpaca Authors
+// Copyright 2019, 2021 The Alpaca Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -15,10 +15,8 @@
 package main
 
 import (
-	"bufio"
 	"encoding/base64"
 	"log"
-	"net"
 	"net/http"
 	"os"
 	"strings"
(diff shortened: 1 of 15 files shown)


samuong:
Most of the changes here are simple renamings of types and functions. The only exception is that I've also merged the `PACData` and `pacData` types. It felt a bit overkill to keep these separate, but @camh- let me know if you think there's a reason this shouldn't be merged?

camh-:
Could you create an issue for this an assign it the right number of story points. It's a little hard to track velocity and burndown without it.

samuong:
I guess I wasn't very clear in my original description, but my plan for the future is to split the `main` package up into smaller packages (with more carefully designed APIs than what's there now). As I do this, I don't want to constantly break compatibility and have to bump the major version over and over again. So I thought it'd be easier to just break compatibility once, and start with a clean slate.

camh-:
> So I thought it'd be easier to just break compatibility once, and start with a clean slate.

I don't think you'll be breaking compatibility at all, since it is all currently in package main (not importable as i understand it). You could just move things bit by bit and as you do that, only export what needs to be exported as you make packages out of things, but up to you.

samuong:
Ah, I hadn't ever actually thought about `main` not being importable, but that makes sense. Just checked and I get an error if I try, even if I import it under a different name. So we don't need this then, I'll abandon this change. Thanks!

PR #65, opened 2021-05-04
Set up golangci-lint

### LINKED ISSUE #64: Set up golangci-lint
As mentioned in https://github.com/samuong/alpaca/pull/62#discussion_r625365032

Instructions here: https://github.com/golangci/golangci-lint-action

samuong:
I also added a Makefile to make it easy to run the linter and coverage checker, as suggested by @juliaogris in #20. It only works on Linux though, so maybe it doesn't belong in the repo?

Fixes #64


pacfinder.go
```diff
@@ -46,7 +46,7 @@ func findPACURLForDarwin() (string, error) {
 	if err := cmd.Start(); err != nil {
 		return "", err
 	}
-	defer cmd.Wait()
+	defer cmd.Wait() // nolint:errcheck
```
camh-, line 49: should be `//nolint:errcheck` - i.e. no space after comment token - these are comments for the machine. not humans. Hard to find definitive info about this, but https://github.com/golang/go/issues/37974#issuecomment-607510041 is a starting point.

Makefile
```diff
@@ -0,0 +1,21 @@
...
+.PHONY: coverage.txt
+coverage.txt:
+	go-acc ./...
+
+coverage.html: coverage.txt
+	go tool cover -html=coverage.txt -o coverage.html
+
+.PHONY: test
+test: coverage.html lint
+	xdg-open coverage.html
```
camh-, line 18: `go tool cover` has logic that works on different OSes to open the coverage output in the browser.

I don't know what `go-acc` is that generates coverage.txt, but what we usually do is:
```
test:
    go test -coverprofile=out/coverage.txt

cover: test
    go tool cover -html=out/coverage.txt
```
See https://github.com/foxygoat/s/blob/92c4b00e14a3576e9dc4ca99b1ca94e0505188f6/Makefile#L25 for an example.

pacfinder.go
samuong: Done.

Makefile
samuong: `go-acc` is another coverage tool that also collects coverage for integration tests (that run across packages). I like having that visibility there, but I guess it is not a standard part of the go toolchain, and I'm not sure if it has a way to open a browser in a cross-platform way.

I'm a bit hesitant to introduce strict test coverage requirements into Alpaca as you've done in the foxygoat/s repo. So I'm going to remove the Makefile from this PR and we can figure it out later.

PR #69, opened 2021-05-24
windows binaries .exe extension to build refinements

jamesmoriarty:
EDIT: I've added two more commits.

Commits: 

1. bugfix: windows binary to include .exe extension that would otherwise make it difficult for users to execute.
2. refactor: release asset naming convention to "alpaca_[version]_[platform]?[.extentsion]"
3. refactor: the build release asset job to use the previous job outputs consistently. 

Screen Shot 2021-05-24 at 8 03 12 pm

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,9 @@ jobs:
   release:
     runs-on: ubuntu-latest
 
+    outputs:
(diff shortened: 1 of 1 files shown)


samuong (approved):
Nice catch, thanks for this!

PR #70, opened 2021-05-24
Allow credentials to be configured via an environment variable

samuong:
This allows non-interactive use of Alpaca on Linux and Windows.


authenticator_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019 The Alpaca Authors
+// Copyright 2019, 2021 The Alpaca Authors
```
camh-, line 1: Whatever you do, don't mention 2020.

main.go
```diff
@@ -53,21 +50,32 @@ func main() {
...
-		fmt.Printf("Password (for %s\\%s): ", *domain, *username)
-		buf, err := term.ReadPassword(int(os.Stdin.Fd()))
-		fmt.Println()
-		if err != nil {
-			log.Fatalf("Error reading password from stdin: %v", err)
-		}
-		a = &authenticator{domain: *domain, username: *username, hash: getNtlmHash(buf)}
-	} else if getCredentialsFromKeyring != nil {
+		src = fromTerminal().forUser(*domain, *username)
+	} else if value, ok := os.LookupEnv("NTLM_CREDENTIALS"); ok {
```
camh-, line 56: I would consider making this:
```
} else if value := os.Getenv("NTLM_CREDENTIALS"); value != "" {
```

That way, you can easily launch alpaca to make it not use the envvar without actually unsetting it in your shell by running:
```
$ NTLM_CREDENTIALS= ./alpaca
```

An empty envvar is as good as an unset envar, so no point even trying to use it I think.

main.go
```diff
@@ -53,21 +50,32 @@ func main() {
...
+	}
+
+	if *printHash {
+		if a == nil {
+			fmt.Println("Please specify a domain (using -d) and username (using -u)")
+			os.Exit(1)
 		}
+		fmt.Printf("# Add this to your ~/.profile (or equivalent) and restart your shell\n")
+		fmt.Printf("NTLM_CREDENTIALS=%q; export NTLM_CREDENTIALS\n",
+			a.username+"@"+a.domain+":"+a.hash)
```
camh-, line 77: perhaps add a `String()` method to `authenticator` that does this formatting?

credentials.go
```diff
@@ -0,0 +1,82 @@
...
+
+func fromEnvVar(value string) *envVar {
+	return &envVar{value: value}
+}
+
+func (e *envVar) getCredentials() (*authenticator, error) {
+	at := strings.IndexRune(e.value, '@')
+	colon := strings.IndexRune(e.value, ':')
+	if at == -1 || colon == -1 || at > colon {
+		return nil, errors.New("invalid $NTLM_CREDENTIALS, please run `alpaca -H`")
```
camh-, line 75: Technically, this bit of code has no idea about envvars or specifically the `NTLM_CREDENTIALS` var. It really is just a string source, which could also be used later when implementing a config file.

Also I wonder whether the parsing of user@domain:hash belongs in `authenticator`, much as I suggested elsewhere that the formatting of the three fields perhaps belongs in a `String()` function on authenticator. At the moment, knowledge of this format is split between here and `main()`.

camh- (approved):
Pretty much LGTM, Just some small comments. Approving in case you don't agree with those comments.

camh- (approved):
LGTM

PR #72, opened 2021-11-14
Pass complete URLs to PAC script, assuming https if scheme is empty

### LINKED ISSUE #58: URLs are not passed to shExpMatch
A `CONNECT` verb used when proxying usually contains just a hostname:port, not a full URL - i.e. it does not contain a scheme.

However, certain code in alpaca expects this to be a URL - in particular, it is passed to `shExpMatch` as a URL but without the scheme, so any matches against a pattern with a scheme will fail. It seems that chrome does pass a URL here, from anecdotal evidence.

I think we need to add the https scheme if it is missing. I'm not sure it makes sense for any scheme other than https to be used for the tunnel.

samuong:
Also change FindProxyForURL() to pass by value rather than by reference. This
function modifies the URL, so make sure to do that on a copy rather than on the
original.

Fixes #58

### DIFF
--- a/pacrunner.go
+++ b/pacrunner.go
@@ -73,9 +73,14 @@ func (pr *PACRunner) Update(pacjs []byte) error {
 	return nil
 }
 
-func (pr *PACRunner) FindProxyForURL(u *url.URL) (string, error) {
+func (pr *PACRunner) FindProxyForURL(u url.URL) (string, error) {
 	pr.Lock()
 	defer pr.Unlock()
+	if u.Scheme == "" {
+		// When a net/http Server parses a CONNECT request, the URL will
+		// have no Scheme. In that case, assume the scheme is "https".
+		u.Scheme = "https"
+	}
 	if u.Scheme == "https" || u.Scheme == "wss" {
 		// Strip the path and query components of https:// URLs.
 		// https://developer.mozilla.org/en-US/docs/Web/HTTP/Proxy_servers_and_tunneling/Proxy_Auto-Configuration_(PAC)_file#Parameters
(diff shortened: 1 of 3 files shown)


pacrunner.go
```diff
@@ -73,9 +73,14 @@ func (pr *PACRunner) Update(pacjs []byte) error {
...
 }
 
-func (pr *PACRunner) FindProxyForURL(u *url.URL) (string, error) {
+func (pr *PACRunner) FindProxyForURL(u url.URL) (string, error) {
 	pr.Lock()
 	defer pr.Unlock()
+	if u.Scheme == "" {
+		// When a net/http Server parses a CONNECT request, the URL will
+		// have no Scheme. In that case, assume the scheme is "https".
+		u.Scheme = "https"
```
camh-, line 82: This could be a little smarter by looking at the port if specified. All the CONNECT examples specify it, but a quick reading of the spec suggests that maybe port is not mandatory. But if port is 80, you could probably safely assume http and not https.

camh- (approved):
One suggestion, but otherwise LGTM

pacrunner.go
samuong: I was considering that, but decided against it because I didn't want to have to maintain a mapping. It would be nice if we could use something like [this services map](https://github.com/golang/go/blob/35a588109b2a6d8b610be08d32aaf99ef1549085/src/net/lookup.go#L36-L55), which is exposed in [net.LookupPort()](https://github.com/golang/go/blob/35a588109b2a6d8b610be08d32aaf99ef1549085/src/net/lookup.go#L406), although that mapping is going in the opposite direction to what we'd need here.

Also iiuc port 80 could mean ws:// too, and maybe that's more likely to be the case when someone does a CONNECT request to port 80?

samuong: Also this: https://gist.github.com/mahmoud/2fe281a8daaff26cfe9c15d2c5bf5c8b

camh-: There is `/etc/services` too.

PR #73, opened 2022-03-27
Change authenticator_test.go from an integration test to a unit test

samuong:
This makes it independent from the code in proxy_test.go, which will
make it easier to move these into separate packages in the future.


camh- (approved):
LGTM (auto-stamp - reviewing tests only is boring)

PR #74, opened 2022-03-27
Run proxy finder function and add result to request context

samuong:
This lets us run the proxy resolution code only once, even if we have to
perform multiple requests (e.g. for authentication).

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -1,4 +1,4 @@
-// Copyright 2019, 2021 The Alpaca Authors
+// Copyright 2019, 2021, 2022 The Alpaca Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -37,7 +37,7 @@ type ProxyHandler struct {
 
 type proxyFunc func(*http.Request) (*url.URL, error)
 
-func NewProxyHandler(proxy proxyFunc, auth *authenticator, block func(string)) ProxyHandler {
+func NewProxyHandler(auth *authenticator, proxy proxyFunc, block func(string)) ProxyHandler {
 	return ProxyHandler{&http.Transport{Proxy: proxy}, auth, block}
 }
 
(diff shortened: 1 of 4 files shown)


proxy.go
```diff
@@ -69,22 +76,17 @@ func (ph ProxyHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
 	id := req.Context().Value(contextKeyID)
-	if err != nil {
-		log.Printf("[%d] Error finding proxy for %v: %v", id, req.Host, err)
-		w.WriteHeader(http.StatusInternalServerError)
-		return
-	}
 	var server net.Conn
-	if u == nil {
+	var err error
+	if proxy, _ := ph.transport.Proxy(req); proxy == nil {
```
camh-, line 82: you're ignoring the error from `ph.transport.Proxy()` - is that to fallback to a direct connection with `net.Dial()` if you get an error with the proxy? Should that at least be logged?

proxy.go
```diff
@@ -69,22 +76,17 @@ func (ph ProxyHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+		server, err = connectViaProxy(req, proxy.Host, ph.auth)
 		var dialErr *dialError
 		if errors.As(err, &dialErr) {
-			log.Printf("[%d] Temporarily blocking unreachable proxy: %q", id, u.Host)
-			ph.block(u.Host)
+			log.Printf("[%d] Temporarily blocking proxy: %q", id, proxy.Host)
+			ph.block(proxy.Host)
 		}
 	}
 	if err != nil {
```
camh-, line 92: On the subject of logging errors, this error does not get logged.

proxy.go
```diff
@@ -37,8 +37,15 @@ type ProxyHandler struct {
 
 type proxyFunc func(*http.Request) (*url.URL, error)
 
-func NewProxyHandler(proxy proxyFunc, auth *authenticator, block func(string)) ProxyHandler {
-	return ProxyHandler{&http.Transport{Proxy: proxy}, auth, block}
+func NewProxyHandler(auth *authenticator, block func(string)) ProxyHandler {
+	getProxyFromContext := func(req *http.Request) (*url.URL, error) {
```
camh-, line 41: Wondering why this is a named closure instead of just a plain function, as it does not close over any variables. You could put the stand-alone function in `proxyfinder.go` alongside the code that adds the value to the context.

camh- (approved):
LGTM.

nit: the commit message misspells "resolution" as "resoultion".

proxy.go
samuong: Yes, that was the original intention. Agree it should be logged, I've just added this.

proxy.go
samuong: Ah you're right - if it goes down the `if` branch, then it won't get logged. But if it goes down the `else` branch, then it'll get logged inside `connectViaProxy()`. To try to keep things a bit consistent, I've added a new function called `connectDirect()` which logs the error, if one occurs.

proxy.go
samuong: It's been a while since I wrote this (it's been sitting on a local branch since last year). I *think* the reason was that it never gets called from anywhere else; it only ever gets called when someone calls the `Transport`'s `Proxy` function. So basically an attempt to minimise the scope of this symbol.

That said, I don't really like how the proxy is added and taken from the context in different middleware layers, so I've gone with your suggestion of moving it into `proxyfinder.go`.

camh- (approved):
LGTM, but it looks like the rebase has gone funny and you've included the commit of the testing PR into this one too somehow.

samuong:
I think this happened when I merged the previous PR and then deleted the branch for that PR. That changed the base branch for *this* PR, leading to the extra commit you saw. I've undeleted that branch for now, and re-set the base commit for this PR, and it looks like it's ok now.

PR #75, opened 2022-03-27
Use net.OpError instead of our own dialError type

samuong:


camh- (approved):
LGTM. I assume that `dialError` actually wasn't being used so that error handling code was dead code before? I can see you've removed `dialError` but there are no diffs that remove anywhere such an error gets created.

samuong:
Yep you're right, `dialError` was something that I added but I never got around to creating the error.

PR #76, opened 2022-03-27
Try to determine PAC URL before downloading, rather than on startup

samuong:
This allows Alpaca to detect changes in the system's PAC URL. Some VPN
clients change the PAC URL to a locally-hosted file upon connection, so
this needs to be checked whenever a network change is detected.


camh- (approved):
LGTM

Issue #11, opened 2019-07-10
Versioning


samuong:
Alpaca releases should be given a version number and this version should be made visible via a `-version` flag and/or printed in the logs.

There is currently no concept of a "release" in Alpaca, so resolving this issue would require a build/release pipeline to be set up.

rtfmoz2:
git describe --tags | head -1 will give you latest release tag on the current cloned repo. If you are using macos you can get the version number with the following variable substitution in KSH.
```
VER=`git describe --tags | head -1`
VER=${VER#v}
VER=${VER%-*-*}
```

rtfmoz2:
I think this indicates that versioning is happening around go modules automatically without there being explicit code - https://go.dev/blog/publishing-go-modules

rtfmoz2:
Got a solution, adding a commit soon

rtfmoz2:
https://github.com/samuong/alpaca/pull/78

PR #77, opened 2022-03-31
Feature/build

rtfmoz2:
Add build versioning for alpaca, this is a precursor for software packaging deployment

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ $ brew install samuong/alpaca/alpaca
 If you want to build on macOS then install latest go and use
 ```
 $ go build -ldflags="-X 'main.Version=v1.1.0' -X 'main.User=$(id -u -n)' -X 'main.Time=$(date)'"
+
 $ ./alpaca -v
 2022/03/31 14:09:49 keyring_darwin.go:88: Found NoMAD credentials for <user> in system keychain
 Version:	 v1.1.0
--- a/README.md
+++ b/README.md
@@ -15,6 +15,16 @@ $ brew tap samuong/alpaca
 $ brew install samuong/alpaca/alpaca
 ```
 
+If you want to build on macOS then install latest go and use
+```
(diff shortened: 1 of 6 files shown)


samuong (commented):
Thanks for contributing this. But I'm a little confused by the purpose of the build scripts here, before we get into the version stuff, can you help me understand this? Comments below.

rtfmoz2:
Oh crap, its showing the stuff I am currently working on, that was not my intention. Those build scripts are unfinished, let me save my latest commits.

rtfmoz2:
Ok there you go... they will make a lot more sense now.  I haven't decided if there is anything to do post-install yet but some man pages would be nice so I have to figure out how to automagically generate some. Hmm looks for docs on alpaca...

samuong:
What you're saying makes sense, and versioning is something I've been thinking about for a while (see #11) but haven't gotten around to yet. There is no version string anywhere in the code, and I'd like to avoid that if possible. The reason is that they get can in the way of merges, or just forgotten about.

I'm currently thinking that the best way to do this is to pass the output of `git describe --tags` to go via ldflags from GitHub Actions and/or look at the `vcs.revision` (and `vcs.modified`) setting in the [BuildInfo](https://pkg.go.dev/runtime/debug#BuildInfo).

I'm not sure exactly what autodocs you'd need; let's continue the conversation over email? I use Gmail and my username there is the same as my GitHub username.

rtfmoz2:
> What you're saying makes sense, and versioning is something I've been thinking about for a while (see #11) but haven't gotten around to yet. There is no version string anywhere in the code, and I'd like to avoid that if possible. The reason is that they can get in the way of merges, or just forgotten about.
> 
> I'm currently thinking that the best way to do this is to pass the output of `git describe --tags` to go via ldflags from GitHub Actions and/or look at the `vcs.revision` (and `vcs.modified`) setting in the [BuildInfo](https://pkg.go.dev/runtime/debug#BuildInfo).
> 
> I'm not sure exactly what autodocs you'd need; let's continue the conversation over email? I use Gmail and my username there is the same as my GitHub username.

I found I can get the version using `git describe --tags | head -1`

PR #79, opened 2022-04-03
BUGFIX: Update to use ioutil

rtfmoz2:
Code to fix go test failure with authenticator_test see issue #80


rtfmoz2:
Ignore this, the WSL version of GO that was installed was 1.13 so the old version of the command has to be used.

PR #81, opened 2022-04-03
FEATURE: Application Version

rtfmoz2:
Solves Issue #11

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -68,7 +68,11 @@ jobs:
         run: |
           echo SDKROOT=$(xcrun --sdk macosx --show-sdk-path) >> $GITHUB_ENV
 
-      - run: go build -v -o ${{ matrix.target.goos }}-${{ matrix.target.goarch }} .
+      - run: |
+          tag=$(git describe --tags | grep '^v' | head -1) 
+          semver=${tag#v}
+          version=${semver%-*-*}
+          go build -v -o ${{ matrix.target.goos }}-${{ matrix.target.goarch }} -ldflags="-X 'main.BuildTime=$(date)' -X 'main.BuildVersion=$version'" .
         env:
           GOOS: ${{ matrix.target.goos }}
           GOARCH: ${{ matrix.target.goarch }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -68,9 +68,7 @@ jobs:
         run: |
           echo SDKROOT=$(xcrun --sdk macosx --show-sdk-path) >> $GITHUB_ENV
 
-      - run: |
-          semver=$(git describe --tags | grep '^v' | head -1) 
-          go build -v -o ${{ matrix.target.goos }}-${{ matrix.target.goarch }} -ldflags="-X 'main.BuildVersion=${semver}'" .
+      - run: go build -v -o ${{ matrix.target.goos }}-${{ matrix.target.goarch }} -ldflags="-X 'main.BuildVersion=${{ needs.release.outputs.version }}'" .
         env:
           GOOS: ${{ matrix.target.goos }}
           GOARCH: ${{ matrix.target.goarch }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -69,10 +69,8 @@ jobs:
           echo SDKROOT=$(xcrun --sdk macosx --show-sdk-path) >> $GITHUB_ENV
 
       - run: |
-          tag=$(git describe --tags | grep '^v' | head -1) 
-          semver=${tag#v}
-          version=${semver%-*-*}
-          go build -v -o ${{ matrix.target.goos }}-${{ matrix.target.goarch }} -ldflags="-X 'main.BuildTime=$(date)' -X 'main.BuildVersion=$version'" .
+          semver=$(git describe --tags | grep '^v' | head -1) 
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -68,7 +68,11 @@ jobs:
         run: |
           echo SDKROOT=$(xcrun --sdk macosx --show-sdk-path) >> $GITHUB_ENV
 
-      - run: go build -v -o ${{ matrix.target.goos }}-${{ matrix.target.goarch }} .
+      - run: |
+          tag=$(git describe --tags | grep '^v' | head -1) 
+          semver=${tag#v}
+          version=${semver%-*-*}
```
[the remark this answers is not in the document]
rtfmoz2: Sounds good, will update the code

version.go
```diff
@@ -0,0 +1,4 @@
+package main
+
+var BuildVersion string
+var BuildTime string
```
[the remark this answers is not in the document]
rtfmoz2: Wasn't entirely sure of the go convention here, happy to do so.

.github/workflows/release.yml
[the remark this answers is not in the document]
rtfmoz2: Do you want to leave the v there as well? v1.2.0-.....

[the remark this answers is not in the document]
rtfmoz2: 3a07b05447f3865e4955593c687e69b087931dac

.github/workflows/release.yml
```diff
@@ -68,7 +68,9 @@ jobs:
         run: |
           echo SDKROOT=$(xcrun --sdk macosx --show-sdk-path) >> $GITHUB_ENV
 
-      - run: go build -v -o ${{ matrix.target.goos }}-${{ matrix.target.goarch }} .
+      - run: |
+          semver=$(git describe --tags | grep '^v' | head -1)
```
samuong, line 72: Since it's got the suffix at the end, it's not really a semver, so we should just call it `version` or something.

version.go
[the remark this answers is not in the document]
rtfmoz2: https://github.com/samuong/alpaca/pull/81/commits/a65e429d4f89c92c54fc45d434fee3989e8632c6

main.go
```diff
@@ -33,14 +33,21 @@ func whoAmI() string {
 }
 
 func main() {
+	var BuildVersion string
```
samuong, line 36: I think this does need to be a global variable rather than a variable that's local to the main function.

After you've moved the variable out, can you test it by building locally (with the appropriate ldflags) and running `alpaca -v`?

main.go
```diff
@@ -33,14 +33,21 @@ func whoAmI() string {
...
 	log.SetFlags(log.LstdFlags | log.Lshortfile)
 	port := flag.Int("p", 3128, "port number to listen on")
 	pacURLFromFlag := flag.String("C", "", "url of proxy auto-config (pac) file")
 	domain := flag.String("d", "", "domain of the proxy account (for NTLM auth)")
+	version := flag.Bool("v", false, "print version number")
```
samuong, line 41: To me, when I see -v I usually think "verbose". What do you think about using -version instead?

Also, I'm nitpicking here, but to me domain and username go together, so it's looks strange that you've put version in between them. Maybe order the variables by the order that we use them, and define version just above port?

samuong (commented):
This is almost there, just a few small comments below.

.github/workflows/release.yml
rtfmoz2: I just realised there is a ${{ needs.release.outputs.version }} in that file. Forest for the trees I swear! We should just use that

main.go
rtfmoz2: Will do

main.go
rtfmoz2: I hear you the problem is -v is usually short for -version.

.github/workflows/release.yml
rtfmoz2: https://github.com/samuong/alpaca/pull/81/commits/2da2d0be24e4d06e5ce6ce08ef3d3101af844015

main.go
rtfmoz2: https://github.com/samuong/alpaca/pull/81/commits/ba054bfb488e9c0010270446e0bdd9bf6c73c640

main.go
rtfmoz2: https://github.com/samuong/alpaca/pull/81/commits/e22f97913e754750cfd396ddc637b0de71ffe338

.github/workflows/release.yml
rtfmoz2: If you want current local commit that is a new requirement. This satisfies the version number requirement for #11

rtfmoz2:
Oh when its merged can you drop a new point release so it has the build number functionality

main.go
```diff
@@ -32,15 +32,23 @@ func whoAmI() string {
 	return me.Username
 }
 
+var BuildVersion string
+
```
samuong, line 36: Just one more small nitpick: can you move this to above the `whoAmI` function? Just wanted to keep all the variables/constants at the top of the file. Other than that, everything here looks good, thanks!

rtfmoz2: https://github.com/samuong/alpaca/pull/81/commits/0221f5d01481fc400685fdb5e70c9d63284c3ed1

samuong (approved):
lgtm, thanks!

rtfmoz2:
Version 1.2.1?

PR #83, opened 2022-04-04
NEW: Build script with commit version

rtfmoz2:
@samuong indicated he would like a version number that included the current commit so he can tell what version a person is testing with. This build script provides this by passing it to the go build command using ldflags.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/bash
+# local build
+version=$(git describe --tags | head -1) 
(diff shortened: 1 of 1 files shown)


samuong:
Hi @rtfmoz2, sorry I missed this earlier. I really consider GitHub Actions to be the build system for Alpaca, and you've already made the change for the Actions config. So I think it's probably not necessary to automate this in multiple places. Let me know if you disagree?

Issue #44, opened 2019-12-22
Support HTTP Negotiate/Kerberos Authentication


samuong:
When Alpaca receives a 407 from an upstream proxy, it just assumes that it's an NTLM proxy and starts to do an NTLM handshake.

In the last few months, I noticed that some of the proxies that I was authenticating to began to send back a "Proxy-Authenticate: Negotiate" header instead of "Proxy-Authenticate: NTLM". This allows the client (Alpaca) to use either NTLM or Kerberos.

Alpaca should check the value of this header, and switch authentication mechanisms accordingly. More information about the Negotiate scheme can be found at https://tools.ietf.org/html/rfc4559 and https://docs.microsoft.com/en-us/dotnet/framework/wcf/feature-details/understanding-http-authentication.

A Kerberos library that might help is https://github.com/jcmturner/gokrb5.

rtfmoz2:
https://support.apple.com/en-au/guide/deployment/depe6a1cda64/web

samuong:
I'm not sure that's correct. According to your link, only iOS and iPadOS (but not macOS) support "Negotiation challenges", and even that only handles "HTTP 401 Negotiate challenges" not HTTP 407 (proxy authentication) challenges. Am I missing something?

rtfmoz2:
If you read the macOS section it says Kerberos is handled at the network
layer. Unless I am missing something.

samuong:
Not sure if you've got a proxy that uses Kerberos auth to test against; if so, does this work for you? When I've tried this in the past Alpaca received a 407 from the upstream proxy and I wasn't able to go any further.

samhaque:
We use Kerberos auth for the proxy, will test and report back if alpaca works

samhaque:
Here is a screenshot of our redacted logs, as you can see we didn't set the NTLM_CREDENTIALS environment variable and curl request to google still worked with our proxy server which supports both NTLM and Kerberos auth.

![image](https://user-images.githubusercontent.com/13789262/163744791-a7488eb0-3b45-426d-ad89-94e1f340fac8.jpeg)

samhaque:
Note I am on a Mac running on Apple silicon, and I compiled the latest version of alpaca for my system from the latest commit. Also my Mac had binded with AD first before I ran Alpaca, without the first bind, the Kerberos ticket wouldn't have generated and the proxy would have failed.

rtfmoz2:
[...]

PR #85, opened 2022-04-20
Log 407 responses from proxy

samuong:
When a proxy server responses with 407 Proxy Authentication Required, Alpaca silently retries the request with authentication. This PR adds logging so that we know when Alpaca has gone down the code path to authenticate to the proxy. This makes it easier to see what's going on in cases such as issue #44.

### DIFF
--- a/authenticator.go
+++ b/authenticator.go
@@ -50,6 +50,7 @@ func (a authenticator) do(req *http.Request, rt http.RoundTripper) (*http.Respon
 		log.Printf("Expected response with status 407, got %s", resp.Status)
 		return resp, nil
 	}
+	log.Printf("2. got %q", resp.Status)
 	challenge, err := base64.StdEncoding.DecodeString(
 		strings.TrimPrefix(resp.Header.Get("Proxy-Authenticate"), "NTLM "))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


camh- (approved):

samhaque:
I spoke with a few users who have requested to set verbosity on the log messages via some environment variable, as they said it's too noisy. 

Also I believe when installed as a brew service the logs are sent to dev null, we can also look at piping the logs to a rotating log file.

samuong:
@samhaque is it too noisy because of this change specifically, or is that more of a general comment?

I'm wondering if it would help to try to rework this PR so that it (usually, unless there's an error somewhere else) prints a single line per request (e.g. `GET http://example.com via "DIRECT", retried with auth`) rather than two separate lines.

samhaque:
So users said noisy in general but that's just how a proxy works, I guess a switch to turn the logging off is what the users are looking for, as there is a performance cost to consider when logging every successful proxy request.

samuong:
Ok, I'll land this as is then, and we can think about how to make it less verbose in a separate issue/pull request. Obviously it's easy enough to just redirect all output to /dev/null, but it would be good to get a better sense of what is still useful and what is noise.

PR #94, opened 2022-05-02
Add support for HTTP and HTTPS proxy resolution schemes

### LINKED ISSUE #92: failed to parse the pac file and obtain the proxy
The pac works in my chrome, but in alpaca, the log shows that :

 proxyfinder.go:128: [1] Couldn't parse proxy: "HTTPS ssl-xxxxxx.yyyyy.com:3389"

Is there anything wrong? Thanks.

samuong:
Fixes #92

### DIFF
--- a/proxyfinder.go
+++ b/proxyfinder.go
@@ -105,27 +105,32 @@ func (pf *ProxyFinder) findProxyForRequest(req *http.Request) (*url.URL, error) 
 	var fallback *url.URL
 	for _, elem := range strings.Split(str, ";") {
 		fields := strings.Fields(strings.TrimSpace(elem))
-		if len(fields) == 1 && fields[0] == "DIRECT" {
+		var defaultPort string
+		if len(fields) == 0 {
+			continue
+		} else if fields[0] == "DIRECT" {
 			log.Printf("[%d] %s %s via %q", id, req.Method, req.URL, elem)
 			return nil, nil
-		} else if len(fields) == 2 && fields[0] == "PROXY" {
-			// The specified proxy should contain both a host and a port, but if for
-			// some reason it doesn't, assume port 80. This needs to be made explicit,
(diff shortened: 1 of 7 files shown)


proxyfinder.go
```diff
@@ -105,27 +105,32 @@ func (pf *ProxyFinder) findProxyForRequest(req *http.Request) (*url.URL, error)
...
-			}
-			if pf.blocked.contains(proxy.Host) {
-				if fallback == nil {
-					fallback = proxy
-				}
-				continue
+		} else if fields[0] == "PROXY" || fields[0] == "HTTP" {
+			defaultPort = "80"
+		} else if fields[0] == "HTTPS" {
+			defaultPort = "443"
```
camh-, line 117: This branch does not result in the proxy scheme being set to https, so it looks like `http.Transport` will end up using http on port 443.

samuong: Yeah ok I see what you mean. When we get something like `HTTPS proxy.example.com` we'll need to establish a TLS connection between Alpaca and proxy.example.com. In the usual case where we just pass the request to http.Transport it should be enough to set the proxy URL scheme to `https`.

But I think we might need to do something different for CONNECT requests. In that case, [connectViaProxy](https://github.com/samuong/alpaca/blob/6f96b6dc9a90d66c99134cd2ac53fd9091e5c4a5/proxy.go#L143) would need to establish a secure tunnel to the proxy, rather than just dial it using net.Dial. Will need to find a bit of time looking into how to do this before I update this PR.

samuong: Ok this is fixed now. Doing so was reasonably straightforward (thanks to [tls.Dial](https://pkg.go.dev/crypto/tls#Dial)) but adding an integration test was a bit more involved. I ended up making one that tests Alpaca against Squid, which can be configured to operate over HTTPS.

Issue #95, opened 2022-05-16
Is it possible to make alpaca listen on all interfaces ?


paulo-dc:
Hi,

Alpaca is a great tool, and replaces easily cntlm on my computer due to its ability to support proxypac.

Unfortunatly I'm facing one issue : I use docker inside wsl2, and because of the fact that alpaca only listens on the localhost interface, I can't use it from a container :(

`tcp        0      0 127.0.0.1:3128          0.0.0.0:*               LISTEN      111/alpaca`

Is there a way to make alpaca listen on all interfaces (`0.0.0.0`) ?

BR

keilin-anz:
Seems like it's currently hardcoded:

https://github.com/samuong/alpaca/blob/master/main.go#L94

But shouldn't be too hard to make it a flag, not sure of the broader implications to the system however

samuong:
It is possible, but I would like to avoid it if we can. The reason is that alpaca authenticates to the upstream proxy on your behalf, so having it listen for external connections would allow other hosts to make use of your proxy credentials. Ideally we'd just open this up to the docker container, and not to other hosts on the local network.

A few options come to mind:

1. Can you run alpaca in the docker container? I think this would be the easiest way to do it, let me know if there's a reason this isn't possible?
2. If not, we could potentially add a flag to somehow just listen on the docker network interface. For example, on my machine, docker's virtual network interface is 172.17.0.1, and alpaca can be made to listen on that address only:

```
[sam@sam-desktop alpaca]$ ip addr show docker0
3: docker0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN group default 
    link/ether 02:42:2c:d4:75:3f brd ff:ff:ff:ff:ff:ff
    inet 172.17.0.1/16 brd 172.17.255.255 scope global docker0
       valid_lft forever preferred_lft forever
    inet6 fe80::42:2cff:fed4:753f/64 scope link 
       valid_lft forever preferred_lft forever
[sam@sam-desktop alpaca]$ git diff
diff --git a/main.go b/main.go
index b1d63c4..a0b75ef 100644
--- a/main.go
+++ b/main.go
@@ -99,7 +99,7 @@ func createServer(port int, pacurl string, a *authenticator) *http.Server {
 
        return &http.Server{
                // Set the addr to localhost so that we only listen locally.
-               Addr:    fmt.Sprintf("localhost:%d", port),
[...]

PR #96, opened 2022-05-25
Add flag to optionally listen on all interfaces

keilin-anz:
Server currently listens on `localhost:${port}` which is fine for most scenarios
This adds the ability to optionally listen on `:${port}` instead (all local IPs)
It's unknown whether this has far reaching implications - very much included just as a jump off point for samuong/alpaca#95 if desired

Potentially addresses samuong/alpaca#95

**NB: previous commits removed since it was a tiny change and clashed with recent merges**

### DIFF
--- a/main.go
+++ b/main.go
@@ -19,9 +19,11 @@ import (
 	"flag"
 	"fmt"
 	"log"
+	"net"
 	"net/http"
 	"os"
 	"os/user"
+	"strconv"
 )
 
 var BuildVersion string
@@ -36,6 +38,7 @@ func whoAmI() string {
 
 func main() {
 	log.SetFlags(log.LstdFlags | log.Lshortfile)
+	host := flag.String("l", "localhost", "address to listen on")
 	port := flag.Int("p", 3128, "port number to listen on")
 	pacurl := flag.String("C", "", "url of proxy auto-config (pac) file")
 	domain := flag.String("d", "", "domain of the proxy account (for NTLM auth)")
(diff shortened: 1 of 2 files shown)


dylanANZ:
I was thinking possibly `-l` and `--listen` which set an address, say `0:0:0:0` or a regular IP.
Default value is still `localhost`

keilin-anz:
> I was thinking possibly -l or --listen which set an address, say 0.0.0.0 or a regular IP.
> Default value is still localhost

Looks like @samuong is on board with that, will update

Amusingly, I was intentionally avoiding that implementation in some misguided attempt to make listening on `:3128` easy - but the proposed approach is definitely much more flexible

main.go
```diff
@@ -36,6 +36,7 @@ func whoAmI() string {
 
 func main() {
 	log.SetFlags(log.LstdFlags | log.Lshortfile)
+	location := flag.String("l", "localhost", "IPv4 address to listen on")
```
samuong, line 39: This isn't just going to accept an "ipv4 address" - it'll accept ipv6 addresses and hostnames too? I'm thinking we should say something like "address to listen on" instead

samuong: Actually this isn't that simple, IPv6 addresses need to be enclosed in square brackets before being passed to [net.Dial](https://pkg.go.dev/net#Dial)...

main.go
```diff
@@ -91,7 +92,7 @@ func main() {
 
 	s := &http.Server{
 		// Set the addr to localhost so that we only listen locally.
-		Addr:    fmt.Sprintf("localhost:%d", *port),
+		Addr:    fmt.Sprintf("%s:%d", *location, *port),
```
samuong, line 95: ...so here, we should probably do something like:

```
Addr: net.JoinHostPort(*location, strconv(*port)),
```

That'll make sure that it gets enclosed in square brackets, if it's an IPv6 address.

keilin-anz:
Resolving rebase conflicts from `main` was pointless, so dropped old commits and redid addressing PR comments

(very lightly) tested with default value (`localhost:3128`), empty value (`:3128`), and an arbitrary local IPv6 address

samuong (approved):
lgtm, thanks @keilin-anz!

PR #102, opened 2022-06-20
Handle empty URLs in output from networksetup

samuong:
On some machines, for some network services, networksetup returns a blank URL instead of `(null)`. This needs to be ignored by Alpaca, otherwise we'll fail to detect PAC URLs that are set in other network services.


camh- (approved):
Seems fine to me

PR #103, opened 2022-06-20
Always check for new PAC files on darwin.

christogav:
This PR suggests a way of using cgo to performantly (< 100uS) pull the auto proxy configuration from the darwin system to ensure always having the current setting.

### DIFF
--- a/main.go
+++ b/main.go
@@ -37,7 +37,7 @@ func whoAmI() string {
 }
 
 func main() {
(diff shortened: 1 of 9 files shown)


samuong (approved):
Sorry Chris, I didn't notice that you'd pushed a new commit since our last chat. This is awesome, thanks for contributing this!

PR #105, opened 2022-09-14
Find and listen on all available networks

marcelocantos:
This was motivated by npm being unable to proxy through alpaca via `npm set config http://localhost:3128` since [node 17+ favours IPv6][node], whereas [Go's `net.Listen` favours IPv4][go] when a hostname is supplied.

  [node]: https://github.com/nodejs/node/issues/40537
  [go]: https://golang.google.cn/pkg/net/#Listen

This PR performs a hostname lookup and detects the networks (tcp4/tcp6) for all returned IPs.


net.go
```diff
@@ -0,0 +1,34 @@
+package main
+
+import (
+	"log"
+	"net"
+)
+
+func networks(hostname string) []string {
+	if hostname == "" {
+		return []string{"net"}
```
samuong, line 10: Should this be "tcp" instead of "net"?

main.go
```diff
@@ -79,11 +79,20 @@ func main() {
...
-	log.Printf("Listening on %s", s.Addr)
-	if err := s.ListenAndServe(); err != nil {
-		log.Fatal(err)
-	}
+	log.Fatal(gogroup(networks(*host), func(network string) error {
+		s := createServer(*host, *port, *pacurl, a)
+		addr := s.Addr
+		if addr == "" {
+			addr = ":http"
+		}
```
samuong, line 87: I don't think line 86 is actually reachable, given that `*port` is an int (can't be an empty string) and we use net.JoinHostPort() (which will always insert a `:`). Do we need this check?

marcelocantos: Good point. I lifte the code directly from `net.ListenAndServer` without thinking too much about it. Fixed.

net.go
marcelocantos: 🤦. Fixed.

Makefile
```diff
@@ -0,0 +1,27 @@
...
+build:
+	go build
+
+# Build all compiling os-arch combinations just to check for regressions. Output
+# binaries get deleted.
+.PHONY: all-osarchs
+all-osarchs:
+	TMPDIR=$${TMPDIR:-/tmp}
+	GOOSNOT="android|ios|js|plan9"; \
+	for dist in $$(go tool dist list | egrep -v "^($$GOOSNOT)/|^darwin/amd64$$"); do \
```
marcelocantos, line 14: I don't know why, but macOS x64 doesn't build, so I had to exclude it.

```
  GOOS=ios GOARCH=amd64
# github.com/samuong/alpaca
./pacfetcher.go:38:14: undefined: pacFinder
./pacfetcher.go:67:15: undefined: newPacFinder
./keyring_darwin.go:55:20: undefined: keychain.NewItem
./keyring_darwin.go:56:29: undefined: keychain.SecClassGenericPassword
./keyring_darwin.go:60:27: undefined: keychain.QueryItem
```

.github/workflows/ci.yml
```diff
@@ -34,9 +34,9 @@ jobs:
 
   test:
     strategy:
-      matrix: 
+      matrix:
         os: [ 'macos-latest', 'ubuntu-latest', 'windows-latest' ]
-        go: [ '1.17' ]
+        go: [ '1.18', '1.19' ]
```
samuong, line 39: Is there a reason you wanted to test both versions rather than just 1.19?

gogroup.go
```diff
@@ -0,0 +1,18 @@
...
+			errch <- err
+		}
+	}
+
+	for _, arg := range args {
+		go gopher(arg)
+	}
+
+	return <-errch
+}
```
samuong, line 18: I could go either way on this, but this feels like a premature abstraction to me - it could just be written directly into main() without having a separate utility function.

If you do decide to leave it here, can you add a copyright/license comment at the top (following the example in other files) and also add your name to AUTHORS?

net.go
```diff
@@ -0,0 +1,34 @@
...
+		}
+	}
+	if ipv4 {
+		nets = append(nets, "tcp4")
+	}
+	if ipv6 {
+		nets = append(nets, "tcp6")
+	}
+	return nets
+}
```
samuong, line 34: Similar to gogroup.go, can you add a copyright/license comment at the top of this file?

.github/workflows/ci.yml
marcelocantos: The idea is to test the earliest supported and the latest.

gogroup.go
marcelocantos: Unrefactored

net.go
marcelocantos: Done

.github/workflows/ci.yml
marcelocantos: If you only want to support the latest, I can remove 1.18.

samuong: Oh ok I see, that's makes sense. And given I'm still on 1.18 on my home PC I'd actually prefer we keep both.

samuong (approved):
lgtm

samuong (approved):
still lgtm, thanks @marcelocantos!

PR #107, opened 2022-12-28
Add distribution packages information

imp-:
Hello,

I took the liberty of packaging alpaca for ArchLinux (in [AUR](https://aur.archlinux.org/packages/alpaca-proxy)) today.
Having a reference to packaged versions in the upstream project seems like a good idea.

This pull request includes a repology badge, which will automatically include other distribution packages if they appear.
As the name 'alpaca' is claimed by multiple projects, I went with 'alpaca-proxy'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,10 @@ $ go get -v -u github.com/samuong/alpaca
 
 Alpaca can be downloaded from the [GitHub releases page][1].
 
+## Install from distribution packages
+
+[![Packaging status](https://repology.org/badge/vertical-allrepos/alpaca-proxy.svg)](https://repology.org/project/alpaca-proxy/versions)
+
 ## Usage
 
 Start Alpaca by running the `alpaca` binary.


samuong:
@imp- sorry for missing this and thanks for setting this up!

I don't really know much about how to package things in AUR, but when new versions of Alpaca are released, how do we make sure the PKGBUILD file is also updated? Would you be ok with mentioning your name in the README against this section, in case Arch users want to contact you about AUR-specific questions?

imp-:
The packager (currently me) is responsible to monitor for new releases and update the PKGBUILD.

The proposed repology badge already provides a [link](https://repology.org/project/alpaca-proxy/versions) containing the [AUR URL](https://aur.archlinux.org/packages/alpaca-proxy), where arch users can interact with the packager.
It seems therefore pointless to duplicate that information here, especially as it might be subject to change (someone else might take over the package etc...).

samuong:
I followed the AUR URL but it looks like I need to create an account before I can flag a package out-of-date or submit a request - and in order to create an account, the registration form requires that I have access to an Arch machine to run a pacman command.

So I don't think I'll be able to contact you directly when I make new releases, but I'm ok to trust that you can do this on your end. I'll just point Arch users to the AUR URL if they ask, thanks.

Also, just a heads up that I'm planning to push out a new release in a few days, to fix another issue that was reported.

imp-:
No worries, problems with packaging are distribution-specific problems after all and it should remain that way.

Thanks for including this pull request!

PR #108, opened 2023-02-15
Remove dodgy connection check

### LINKED ISSUE #106: Feature request: Direct mode
Hi!
Thanks for this great project!

It would be nice to have the feature to start alpaca in a mode where all requests are routed as "direct". This way I could always keep the proxy configured and start alpaca with this flag when I'm connected to a network without a proxy.

samuong:
This behaviour worked well on some corporate networks, but breaks on others. From now on, users who use a custom PAC file with a file:// URL will need to do this check themselves, before returning a PROXY directive.

Fixes #106

### DIFF
--- a/pacfetcher_test.go
+++ b/pacfetcher_test.go
@@ -15,7 +15,6 @@
 package main
 
 import (
-	"net"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
@@ -125,14 +124,8 @@ func TestPacFromFilesystem(t *testing.T) {
 	pacPath := path.Join(tempdir, "test.pac")
 	require.NoError(t, os.WriteFile(pacPath, content, 0644))
 	pacURL := &url.URL{Scheme: "file", Path: filepath.ToSlash(pacPath)}
-
 	pf := newPACFetcher(pacURL.String())
-	pf.monitor = &netMonitorImpl{
-		getAddrs: func() ([]net.Addr, error) {
-			return []net.Addr{&net.IPAddr{IP: net.ParseIP("127.0.0.1")}}, nil
-		},
-	}
-
+	pf.monitor = newNetMonitor()
 	assert.Equal(t, content, pf.download())
 	assert.True(t, pf.isConnected())
 }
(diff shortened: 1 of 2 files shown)


samhaque:
Hi, could we please get this reviewed and merged it's blocking a lot of developers who rely on making custom pac files

samhaque:
Looks like all CI checks passed, could you have a look @camh- @juliaogris @marcelocantos @samuong

pacfetcher_test.go
```diff
@@ -149,18 +126,13 @@ func TestPacFromFilesystem(t *testing.T) {
...
 	pacURL := &url.URL{Scheme: "file", Path: filepath.ToSlash(pacPath)}
 
-	tn := testNetwork{false}
 	pf := newPACFetcher(pacURL.String())
 	pf.monitor = &netMonitorImpl{
-		getAddrs: func() ([]net.Addr, error) { return tn.InterfaceAddrs() },
-	}
-	pf.lookupAddr = func(ctx context.Context, addr string) ([]string, error) {
-		return tn.LookupAddr(ctx, addr)
+		getAddrs: func() ([]net.Addr, error) {
```
camh-, line 131: Is this custom monitor needed for this test? I guess if the tests were to run on a host with no network interfaces (i.e. `net.InterfaceAddrs` returns an empty slice) then we may not trigger a download, but there is a `fakeNetMonitor` in this file which I assume was for this purpose. But is it even conceivable that the host running the test has no network interfaces at all? If not, the default netMonitor should suffice for this test.

If I've got this totally wrong, perhaps a comment for why there's a custom `getAddrs` function might help?

camh- (approved):
LGTM

pacfetcher_test.go
samuong: I think you're right, the `getAddrs` function will only be called once, and it doesn't really matter what it returns, as long as it returns _something_. It might be theoretically possible to run this test on a host with no network interfaces, but we have a number of other unit tests that rely on having a local interface and network stack (e.g. the tests in `proxy_test.go` create local servers on various ports).

camh- (approved):

Issue #118, opened 2024-05-13
NTLM http response error , failure to read HTTPResponse is not handling malformed bluecoat http response special characters 


ziyiwang:
hi ,we have some NTLM responses coming back from our upstream bluecoat proxies

and the response is not accepted by alpaca , in this module 
2024/03/22 15:34:14.930490 proxy-go:168: [] Got2 "malformed HTTP response \"<HTML><HEAD>\"" response, retri ed with auth in proxy.go page full err object is null?

in transport.go

func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {

can we split this function
	return http.ReadResponse(t.reader, req)

into separeate sendrequest , get response etc, so we can get the response back first before deciding its a malformed http response

ziyiwang:
i found the issue, its related to the extra httpresponse body bluecoat is sending back with ntlm auth challenge, i will submit a PR soon

PR #119, opened 2024-05-23
Update transport.go , to fix ntlm 407 challenge httpresponse containing extra httpbody issue

ziyiwang:
to fix ntlm 407 challenge httpresponse containing extra httpbody issue issue #118 

extra httpBody in httpResponse with 407 , is not read and stuck in stream, the body needs to be read and discarded before next HTTP response comes in (HTTP 200) , otherwise the stuck body in stream will cause HTTP response error

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -21,6 +21,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"io"
 )
 
 // transport creates and manages the lifetime of a net.Conn. Between the time that a remote server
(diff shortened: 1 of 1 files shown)


transport.go
```diff
@@ -57,7 +58,24 @@ func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	if err := req.Write(t.conn); err != nil {
 		return nil, err
 	}
-	return http.ReadResponse(t.reader, req)
+	var resp2 *http.Response
+	var err2 error
+
+	resp2,err2 = http.ReadResponse(t.reader, req)
```
samuong, line 64: I'm not sure it's correct to read another response here. This function gets called by [ProxyHandler.proxyRequest()](https://github.com/samuong/alpaca/blob/4e40b40f48286fce494e9ec752915e64d3009f94/proxy.go#L190), and the body of the request gets closed [later in that file](https://github.com/samuong/alpaca/blob/4e40b40f48286fce494e9ec752915e64d3009f94/proxy.go#L206). Since this is done in a `defer` statement, which I think is the bug that you're running into.

If that's the case, then I think #120 would fix it - any chance you can give that branch a try and let me know if it fixes your issue?

samuong (commented):
Thanks @ziyiwang for the PR, and for the bug report! These look like two separate bugs that you're running into, I've left a comment on each of the proposed fixes.

transport.go
ziyiwang: sure let me test it out and come back to you, most likely removing defer will also fix the issue

PR #120, opened 2024-05-26
Close response body before moving on to next request/response

### LINKED ISSUE #118: NTLM http response error , failure to read HTTPResponse is not handling malformed bluecoat http response special characters

samuong:
Fixes #118

### DIFF
--- a/authenticator.go
+++ b/authenticator.go
@@ -50,6 +50,7 @@ func (a authenticator) do(req *http.Request, rt http.RoundTripper) (*http.Respon
 		log.Printf("Expected response with status 407, got %s", resp.Status)
 		return resp, nil
 	}
+	resp.Body.Close()
 	challenge, err := base64.StdEncoding.DecodeString(
 		strings.TrimPrefix(resp.Header.Get("Proxy-Authenticate"), "NTLM "))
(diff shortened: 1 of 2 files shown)


ziyiwang:
sure let me test it out and come back to you, most likely removing defer  will also fix the issue

ziyiwang:
No problem, thanks, yup agree with proper handling close body, should resolve issue, i tested with a free users on our end with the io stream fix, also confirms fix issue as body from 407 is not really needed,  bit i leave it in your capable hands, thanks mate,  great project, really address a gap we have with cntlm

PR #138, opened 2025-01-20
Use Ubuntu 22.04 arm64 runners to build on Linux arm64

samuong:


camh-:
I was just replying on the other now-merged PR about how to get cross-compiling going and was going to add a linux/arm64 build to CI but I see it is already there, and now I see this PR that updates it. What I was going to post was:

Cross compiling should not be too hard - you just need the arm64 compiler installed and to set `CC=<cross-compiler>` during the build.

The runners you are running on are Ubuntu 20.04, so installing the apt package `gcc-10-aarch64-linux-gnu` in a setup step in the matrix for the linux/arm64 build and setting `CC=aarch64-linux-gnu-gcc-10` in the environment should make it work. You may also want a comment to update this version if you update the version of the ubuntu runner.

But I decided rather than just writing about it, I'd do it. https://github.com/camh-/alpaca/commit/7b4a1d5ebe90f52984fc948d54c84770db600f49 is the result which does not require an arm runner. Feel free to take that if you prefer to use amd64 runners if contention for the arm64 runners gets too much.

samuong:
Hi Cam, thanks for that! Actually I made a mistake, there isn't any contention for runners, this was just due to a copy and paste error on my part (explained below). So I'm planning to use the arm64 runners, but I appreciate the tip on how to do cross compilation.

The first commit in this PR (1d14f794bdfdb4767d237f8538fc9ee12f3785c2) uses the `ubuntu-20.04-arm` runner (which matches the Ubuntu version we use for amd64) but GitHub only supports `ubuntu-22.04-arm` and `ubuntu-24.04-arm`. I changed this to `ubuntu-22.04-arm` in ef4031e7edb14142d7a3f7938dfa61521dd0f339 and it ran in 23s, which is even faster than the Mac and Windows runners.

PR #141, opened 2025-03-02
Implement myIpAddressEx()

samuong:


pacrunner.go
```diff
@@ -202,17 +200,52 @@ func myIpAddress(call otto.FunctionCall) otto.Value {
...
+func myIpAddressEx(_ otto.FunctionCall) otto.Value {
+	// https://chromium.googlesource.com/chromium/src/+/ee43fa5328856129f46566b2ea1be5811739681c/net/docs/proxy.md#resolving-client_s-ip-address-within-a-pac-script-using-myipaddressex
+	public := []string{"8.8.8.8", "2001:4860:4860::8888"}
+	if ips := probeRoutes(public); ips != "" {
+		return toValue(ips)
+	}
+	if ips := resolveHostname(true); len(ips) > 0 {
+		var b strings.Builder
+		for _, ip := range ips {
+			b.WriteRune(';')
```
camh-, line 212:
```
for i, ip := range ips {
    if i > 0 {
        b.WriteRune(';')
    }
```

It's pretty simple to not put the leading `;` on in the first place, removing the need to strip it off again.

pacrunner.go
```diff
@@ -202,17 +200,52 @@ func myIpAddress(call otto.FunctionCall) otto.Value {
...
+		if localAddr == "" {
+			continue
+		}
+		if _, ok := set[localAddr]; ok {
+			continue
+		}
+		set[localAddr] = struct{}{}
+		slice = append(slice, localAddr)
+	}
+	return strings.Join(slice, ";")
```
camh-, line 236: If the order of the output does not need to match the order of the input, you don't really need the slice and you could simplify this:

```
    set := map[string]struct{}{}
    for _, address := range addresses {
        if localAddr := probeRoute(address); localAddr != "" {
            set[localAddr] = struct{}{}
        }
    }
    return strings.Join(slices.Collect(maps.Keys(set)), ";")
```

This does need Go 1.23 for `slices.Collect()` and `maps.Keys()` though.

camh- (approved):
LGTM. Just some little nitpicks suggestions.

pacrunner.go
samuong: I was still on 1.22 but yes this would be nicer. I still need to check but I'll do it this way if I can confirm that order isn't important.

pacrunner.go
samuong: Ok, my original thinking was that trimming the leading `;` would save having to do the check inside the loop. But I'll do it this way so that people reading it don't think it's a bug, at least initially.

samuong: I used strings.TrimPrefix() because it handles the `len(ips) > 0` case, but as you've pointed out that's never going to happen :facepalm:

pacrunner.go
samuong: I haven't found out whether order is important or not, but I think I'll just leave this as is, just in case it is.

Issue #144, opened 2025-04-14
MacOS - No PAC URL specified or Detected


Enelass:
When the PAC file URL is set in the Network Interface / Proxies / Automatic proxy configuration on MacOS, Alpaca will read it and use it.
However it this isn't set, Alpaca will resort to making direct connections only...
Even if `scutil --proxy` does define the PAC URL...

This is because Alpaca is set to pull the value from `getPACUrlFromSCDynamicStoreCopyValue` and is omitting `getPACUrlFromSCDynamicStoreCopyProxies`

Enelass:
### Added Redundant PAC URL Detection Methods in https://github.com/Enelass/alpaca_pacfinder_darwin_fix

**Before:**
- Single method for PAC URL detection using SCDynamicStoreCopyValue

**After:**
- Two methods for PAC URL detection:
  1. Original method using SCDynamicStoreCopyValue
  2. New method using SCDynamicStoreCopyProxies
- Fallback mechanism: if the first method fails, the second is tried

### 3. Code Structure Improvements

**Before:**
- Single monolithic function for PAC URL detection
- Limited comments and documentation

**After:**
- Split functionality into separate methods with clear responsibilities
- Better function naming (getPACUrlFromSCDynamicStoreCopyValue, getPACUrlFromSCDynamicStoreCopyProxies)
- Added commented logging statements (currently disabled) for debugging

Enelass:
Awaiting PR approval: https://github.com/samuong/alpaca/pull/147/files

Enelass:
Sorry for the delay in fixing this. PR now at https://github.com/samuong/alpaca/pull/154

PR #147, opened 2025-06-02
Fix issue #144: Update pacfinder_darwin.go to check PAC URL in SCDynamicStoreCopyProxies as fallback

Enelass:
### 1) Added Redundant PAC URL Detection Methods in `pacfinder_darwin.go`

**Before:**
- Single method for PAC URL detection using SCDynamicStoreCopyValue

**After:**
- Two methods for PAC URL detection:
  1. Original method using SCDynamicStoreCopyValue
  2. New method using SCDynamicStoreCopyProxies
- Fallback mechanism: if the first method fails, the second is tried

PR pacfinder_darwin fix

### 2) pacfinder Test Files deletion: `pacfinder_unix_test` and `pacfinder_darwin_test`
Remove GO test files for pacfinder as these fixes weren't of good quality:

- Hardcoded PAC url or `http://internal.anz.com/proxy.pac` for ANZ an Australian bank.
- And resorting to using `bash` command `scutil --proxy` as opposed to the C functions for MacOS System Config 

More info here: https://github.com/Enelass/alpaca_pacfinder_darwin_fix/blob/master/Test_Implementation_Issues.md


Enelass:
Morning @samuong ,
Are you OK reviewing this PR please ?
Let me know if you have any questions.
 
 Regs, Flo

pacfinder_darwin_test.go
samuong, line 1: I think I agree with you on the whole that this there are a lot of improvements that can be made to these tests. But until we make those improvements, is there any reason not to keep these tests?

Especially the pacfinder_unix_test.go file, which shouldn't be impacted by your change. Even though it's not ideal, it does provide a very basic check that everything hangs together as it should.

I agree that it's a bit dodgy to use an "internal ANZ" URL as a test but this is really just an arbitrary string, there's no such actual URL inside ANZ and the test does not rely on this URL existing. The reliance on bash is a bit of a holdover from before we moved over to calling the native macOS functions via CGo.

pacfinder_darwin.go
```diff
@@ -20,104 +19,166 @@ package main
...
+	if pacUrl != "" {
+		// log.Printf("Using PAC URL from SCDynamicStoreCopyValue method: %s", pacUrl)
+		return pacUrl, nil
+	}
 
-func (finder *pacFinder) pacChanged() bool {
-	if url, _ := finder.findPACURL(); finder.pacUrl != url {
-		finder.pacUrl = url
-		return true
+	// Second method: Using SCDynamicStoreCopyProxies
```
samuong, line 73: Is there a reason you wanted to maintain SCDynamicStoreCopyValue and use SCDynamicStoreCopyProxies only as a fallback?

I'm not super familiar with these APIs but ChatGPT seems to think the latter is the higher-level API and will handle things like picking from multiple network interfaces, and different user configuration profiles, that the lower-level API will not.

samuong (commented):
Hi Flo, thanks for putting this together, and apologies for missing this when you originally sent it. I have a few questions below.

pacfinder_darwin_test.go
Enelass: Yes happy for you to keep these as if.
I thought the URL was the actual real one for ANZ which would be deemed minor information disclosure. Also scutil is not as consistent with using CGo

Enelass:
CLOSING as new PR is sent at https://github.com/samuong/alpaca/pull/154
@samuong

PR #150, opened 2025-06-21
fix failure to download PAC from file URL with Windows drive letter

basinilya:
- Replaced hardcoded NewFileTransport(http.Dir("C:")) with the fs root from the URL itself. This allows UNC paths and drive letters other than C:. This also allows drive letters within the URL (previously it was trying to open C:some/path). Paths starting with a single slash still work and are resolved relative to the current working drive.
- Prefixed URL.Path=C:/some/path with a slash to prevent C: being parsed as host name. This is fixed in TestPacFromFilesystem and when the program argument is a relative URI
- Changed how relative paths can be passed. Previously only dot was working because of the special handling in the File Transport and you had to specify the file: scheme anyway. Now the relative URI must contain no file: scheme and it's converted to an absolute file URL when the program starts

### DIFF
--- a/pacfetcher.go
+++ b/pacfetcher.go
@@ -20,6 +20,8 @@ import (
 	"io"
 	"log"
 	"net/http"
+	"net/url"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"time"
@@ -33,10 +35,10 @@ const maxResponseBytes = 1 * 1024 * 1024
 var delayAfterFailedDownload = 2 * time.Second
 
 type pacFetcher struct {
-	pacFinder  *pacFinder
-	monitor    netMonitor
-	client     *http.Client
-	connected  bool
+	pacFinder *pacFinder
+	monitor   netMonitor
+	client    *http.Client
+	connected bool
 	//cache  []byte
 	//modified time.Time
 	//fetched time.Time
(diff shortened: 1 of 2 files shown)


samuong:
Ok, yeah I looked into this and there's a proposal to add functions to convert between urls and paths in https://github.com/golang/go/issues/32456. But it's not done yet.

It would be great if you could add some tests, and we can separately keep an eye on that proposal to see if the new functions become useful in the future.

I thought it might be nice to have tests that span across both C: and another drive, but if you don't think it's useful, that's fine. I wasn't planning on merging `TestOnWindowsWithSubst()` as it was just an example, but I did notice that some of the runner images were no longer available, so I've repurposed #155 to fix that.

CC @alexbrainman in case he's still interested in Windows paths :)

basinilya:
* Extracted two functions so they can be tested without the actual request: fixFileRequestUrl and extractNativePath.
* Added unit tests for fixFileRequestUrl, extractNativePath, validateFileUri, and newPACFetcher

samuong:
@alexbrainman Thanks for the pointers! Probably best not to wait for the standard library to offer these functions then.

@basinilya Sorry I know this is taking a really long time (I'm trying to juggle it with other things). But we should make sure we're on the same page about how this code should work, before we get too far into the weeds of code review. So a few questions:

1. For absolute paths, should we use the Go implementation (just copy the file over as Alex suggested)?
2. Do we need to have support for relative paths? It looks like they don't support this, is it a must-have feature for what you're trying to do? I'm wary of adding functionality which will need to be supported, especially if we find edge cases in the future, such as...
3. Should we be implementing support for the legacy/unhealthy file:// URI syntax (see below for what this means)? If we don't, it won't be compatible with Internet Explorer.

As I was looking into file:// URI support, I noticed a few interesting (maybe just to me) things:

- As of IE11, file:// URIs for PAC scripts are "[deprecated](https://learn.microsoft.com/en-au/archive/blogs/ieinternals/understanding-web-proxy-configuration#note-file-based-proxy-scripts-deprecated)" (actually it looks like it's disabled and can only be re-enabled by changing a registry key)
- Even if it's re-enabled, IE requires you to use an "[unhealthy syntax](https://learn.microsoft.com/en-au/archive/blogs/ieinternals/understanding-web-proxy-configuration#note-file-based-proxy-scripts-require-unhealthy-syntax)" for file:// URIs. The link to the page that describes this syntax is broken, but I found an [archived copy](https://web.archive.org/web/20131015233714/http://blogs.msdn.com/b/freeassociations/archive/2005/05/19/420059.aspx) (but note that on this page it's referred to as the "legacy" syntax, rather than the "unhealthy" syntax)
- If Edge hasn't changed the Chromium implementation, then it [won't allow file:// URIs](https://github.com/chromium/chromium/blob/35d4428647813febdb447541efcf5a2e4a9979a0/net/proxy_resolution/pac_file_fetcher_impl.cc#L326-L333) or [redirects to them](https://github.com/chromium/chromium/blob/35d4428647813febdb447541efcf5a2e4a9979a0/net/proxy_resolution/pac_file_fetcher_impl.cc#L237-L239)

For the three questions above, I'm leaning towards no, yes and yes (and detecting legacy syntax by checking whether the separator is a forward or back slash). But I'm open to changing my mind on all of these, if you have reasons for wanting it to work a certain way?

basinilya:
>     1. For absolute paths, should we use the Go implementation (just copy the file over as Alex suggested)?

The Go implementation additionally supports explicit `host="localhost"` in file URL and a UNC host like in `file://somehost/path` -> `\\somehost\path`.
My implementation can normalize slashes like java and browsers do, but Go can't.
I think the latter is more important. Support for non-empty host can be added separately.

>     2. Do we need to have support for relative paths?

For portable distributions for Windows, yes. Then the pac file can be put to the folder with alpaca.exe.

>     3. Should we be implementing support for the legacy/unhealthy file:// URI syntax (see below for what this means)? If we don't, it won't be compatible with Internet Explorer.

Yes, maybe. it can be done separately. For now, the recommended way to obtain a file:// URL for Alpaca should be this: "drag your pac file into a modern browser (not IE) and then copy the resulting URL from the address bar".

samuong:
Hi @basinilya, thanks for your answers, I wasn't aware of the need to support a host of localhost. The portable distributions on Windows is an interesting use case as well, and it would be good to support that.

When you say "normalize slashes", did you mean convert between forward- and back-slashes, or did you mean canonicalisation (e.g. `//` -> `/`)?

The general idea behind Alpaca was that if your system is configured with a PAC URL, then your command line apps should "just work" as well (the only requirement being that you set `http_proxy` and/or `https_proxy` to point to Alpaca of course). There are two use cases for file:// URIs that I can think of:

1. A user has a file:// URI configured in their system's proxy settings, and Alpaca needs to detect that value and use it. I think this is very unlikely (since Chrome and Edge don't support it) but if it is the case, then they probably had to jump through weird hoops to get it to work with Internet Explorer, which means we need to support the "unhealthy syntax". It would be nice to support this, but it would also be ok to just say we don't (like Chrome and Edge have).
2. A user just wants to run Alpaca with some different settings, maybe they've got a customised PAC file, or they're using a packaged "portable" executable which is distributed with a PAC file.

In the latter case, I'm wondering support for `data:` URIs would suffice? On a Unix system, a user could do something like `./alpaca -C "data:base64,$(base64 proxy.pac)"`. On Windows/PowerShell (according to Gemini; I haven't tried this myself) they could run:

```
.\alpaca -C "data:base64,$( [System.Convert]::ToBase64String((Get-Content -Path proxy.pac -Encoding Byte)) )"
```

This could be wrapped up in a PowerShell script to make it easy for end users. Do you think this approach would be too onerous for users?

I'm actually keen to deprecate and remove file:// URI support, as there are so many edge cases and details that differ between implementations (e.g. Go, Java, Chrome, Firefox, IE, Edge). But doing so requires that we have a suitable alternative in place first.

basinilya:
> When you say "normalize slashes", did you mean convert between forward- and back-slashes, or did you mean canonicalisation (e.g. `//` -> `/`)?

The latter

> In the latter case, I'm wondering support for `data:` URIs would suffice?

pac files can be quite big

> there are so many edge cases and details that differ between implementations

I think you're dramatizing

samuong:
Hah, I get you'd feel that way, but you'd be surprised at how many support requests I'm getting from Windows users. These are quite difficult to deal with at the moment especially since I don't have a Windows machine of my own. So I hope you'll understand that I'd like to keep this code path as simple as possible.

One other approach I think would work well is if we just disallowed file:// URIs but accepted paths to the `-C` flag. And then we just read the file using `os.ReadFile()`, and skip the whole `http.fileTransport` and path mangling thing. If you agree and are interested, please let me know.

samuong:
Hi @basinilya, I'm going to close this one off, but you might be able to do what you want once #157 lands.

PR #152, opened 2025-07-07
Fix: dual stack network not working

LucasSnatiago:
Why not set dual stack IPv4 and IPv6 by default? 
In that cause both protocols can coexist and the user can choose the one that it wants the most. The current implementation works as a tunnel as well. Example:

My Package -> Proxy: https://[::1]:3128 -> Network: some IPv4

### DIFF
--- a/main.go
+++ b/main.go
@@ -38,8 +38,6 @@ func whoAmI() string {
 
 type stringArrayFlag []string
 
-var hosts stringArrayFlag
-
 func (s *stringArrayFlag) String() string {
 	return fmt.Sprintf("%v", *s)
 }
@@ -54,6 +52,8 @@ func (s *stringArrayFlag) Set(value string) error {
 
 func main() {
 	log.SetFlags(log.LstdFlags | log.Lshortfile | log.Lmicroseconds)
+
+	var hosts stringArrayFlag
 	flag.Var(&hosts, "l", "address to listen on")
 	port := flag.Int("p", 3128, "port number to listen on")
 	pacurl := flag.String("C", "", "url of proxy auto-config (pac) file")
--- a/main.go
+++ b/main.go
@@ -36,9 +36,25 @@ func whoAmI() string {
 	return me.Username
 }
 
+type stringArrayFlag []string
+
+var hosts stringArrayFlag
+
+func (s *stringArrayFlag) String() string {
+	return fmt.Sprintf("%v", *s)
+}
+
+func (s *stringArrayFlag) Set(value string) error {
+	if value == "" {
+		return nil
+	}
+	*s = append(*s, value)
+	return nil
+}
+
 func main() {
 	log.SetFlags(log.LstdFlags | log.Lshortfile | log.Lmicroseconds)
-	host := flag.String("l", "localhost", "address to listen on")
+	flag.Var(&hosts, "l", "address to listen on")
 	port := flag.Int("p", 3128, "port number to listen on")
 	pacurl := flag.String("C", "", "url of proxy auto-config (pac) file")
 	domain := flag.String("d", "", "domain of the proxy account (for NTLM auth)")
@@ -46,6 +62,11 @@ func main() {
 	printHash := flag.Bool("H", false, "print hashed NTLM credentials for non-interactive use")
 	version := flag.Bool("version", false, "print version number")
 	flag.Parse()
+
+	// default to localhost if no hosts are specified
+	if len(hosts) == 0 {
+		hosts = append(hosts, "localhost")
+	}
 
 	if *version {
 		fmt.Println("Alpaca", BuildVersion)
@@ -82,18 +103,20 @@ func main() {
 
 	errch := make(chan error)
 
-	s := createServer(*host, *port, *pacurl, a)
+	for _, host := range hosts {
+		s := createServer(host, *port, *pacurl, a)
 
-	for _, network := range networks(*host) {
-		go func(network string) {
-			l, err := net.Listen(network, s.Addr)
-			if err != nil {
-				errch <- err
-			} else {
-				log.Printf("Listening on %s %s", network, s.Addr)
-				errch <- s.Serve(l)
-			}
-		}(network)
+		for _, network := range networks(host) {
+			go func(network string) {
+				l, err := net.Listen(network, s.Addr)
+				if err != nil {
+					errch <- err
+				} else {
+					log.Printf("Listening on %s %s", network, s.Addr)
+					errch <- s.Serve(l)
+				}
+			}(network)
+		}
 	}
 
 	log.Fatal(<-errch)
--- a/main.go
+++ b/main.go
@@ -1,4 +1,4 @@
-// Copyright 2019, 2021, 2022 The Alpaca Authors
+// Copyright 2019, 2021, 2022, 2025 The Alpaca Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


samuong:
The reason Alpaca listens on `localhost:3128` (by default) is because I wanted to make sure that the proxy was only available to processes running on the local host.

If we listen to `:3128` instead, Alpaca will accept requests from other hosts too. This is probably not what users (or at least their proxy administrators) want. It shouldn't allow employees to send HTTP requests to their coworkers' instances of Alpaca, since Alpaca is configured with their NTLM credentials and will issue requests on that user's behalf.

By default, the networks() function will do a DNS lookup of `localhost` which should return both `127.0.0.1` and `::1` if the machine is set up for both IPv4 and IPv6. Is this not working for what you're trying to do?

samuong:
It depends on how your system is configured. A few things you could check:

1. What does `net.LookupIP("localhost")` return on your system?
2. Does this match the output of `grep localhost /etc/hosts`?
3. And what does `net.InterfaceAddrs()` return?
4. And does this match the output of `ip addr` (on Linux) or `ifconfig -a` (on Mac) or I think `ipconfig` (on Windows)?

LucasSnatiago:
```go
package main

import (
	"fmt"
	"net"
)

func main() {
	host, _ := net.LookupIP("localhost")
	fmt.Println("1 - net.LookupIP(\"localhost\") = ", host)
	// 1 - net.LookupIP("localhost") =  [127.0.0.1]

	addr, _ := net.InterfaceAddrs()
	fmt.Println("3 - net.InterfaceAddrs() = ", addr)
	// 3 - net.InterfaceAddrs() =  [127.0.0.1/8 10.107.2.16/24 172.18.0.1/16 172.17.0.1/16 ::1/128 fe80::967c:d727:2352:f6a6/64]
}
```

2 -
```
127.0.0.1       localhost
::1     ip6-localhost ip6-loopback
```
What do you think of using: ```strings.HasSuffix()``` instead of ```if strings.Compare(hostname, "localhost") == 0 || hostname == "" {``` to get both 127.0.0.1 and ::1. It would fix my problem of IPv6 not working.

4 - Yes, it matches the command on my wsl.

samuong:
That's interesting, what OS are you using?

From what I can tell, Debian and its derivatives set up `localhost` to map to only `127.0.0.1` and have `ip6-localhost` map to `::1`, whereas other Linux distros as well as macOS (and Windows I think) have `localhost` map to both loopback addresses.

I'm not sure how a suffix match would work - what value would you use for the first argument to `strings.HasSuffix()`? Alpaca doesn't loop over every entry in the hosts file, so it doesn't check that `ip6-localhost` even exists unless asked explicitly.

One option could be to modify Alpaca to allow the `-l` flag to take multiple values. There's an example of how to do this (example 3) in https://pkg.go.dev/flag#example-package. Then you'd need to launch Alpaca with something like `alpaca -l localhost -l ip6-localhost ...`. What do you think?

LucasSnatiago:
I'm using windows 11 with WSL 2:
```
wsl --version
WSL version: 2.5.9.0
kernel version: 6.6.87.2-1
WSLg version: 1.0.66
MSRDC version: 1.2.6074
Direct3D version: 1.611.1-81528511
DXCore version: 10.0.26100.1-240331-1435.ge-release
Windows version: 10.0.26100.3775
```
winver

Sorry for the Portuguese version of the OS, but the only things that matter is the Windows 11 logo with the compilation number. I cannot modify the language of my system.


And using -l 2 times sound a nice idea actually. You can close this pr if wanted and add it to the README should be enough for all the next people using this tool.

LucasSnatiago:
I'm using my own code right now that solved my problem (GoProxy)[https://github.com/LucasSnatiago/GoProxy] it is way simpler than your code and less reliable hahaha. So, whenever I have some spare time I'll change it, but for now I'm quite busy as well.

LucasSnatiago:
Support for listening on multiple hosts added. Can you give a look at it?

main.go
```diff
@@ -82,18 +103,20 @@ func main() {
...
-				log.Printf("Listening on %s %s", network, s.Addr)
-				errch <- s.Serve(l)
-			}
-		}(network)
+	for _, host := range hosts {
+		s := createServer(host, *port, *pacurl, a)
+
+		for _, network := range networks(host) {
+			go func(network string) {
+				l, err := net.Listen(network, s.Addr)
```
samuong, line 111: I noticed that since you're calling `createServer()` multiple times now, this will create multiple instances of the entire middleware chain. This will lead to some undesired behaviour, such as log correlation ids being duplicated, pac files being downloaded multiple times, etc.

For example, I noticed the "No PAC URL specified or detected" line appears twice:

```
[sam@sam-laptop alpaca]$ ./alpaca -l 127.0.0.1 -l ::1
2025/07/26 16:14:17.914845 main.go:90: Credentials not found, disabling proxy auth: cannot get user secret from keyring: secret not found in keyring
2025/07/26 16:14:17.921910 pacfetcher.go:94: No PAC URL specified or detected; all requests will be made directly
2025/07/26 16:14:17.922537 main.go:115: Listening on tcp4 127.0.0.1:3128
2025/07/26 16:14:17.927779 pacfetcher.go:94: No PAC URL specified or detected; all requests will be made directly
2025/07/26 16:14:17.927956 main.go:115: Listening on tcp6 [::1]:3128
2025/07/26 16:14:43.132761 proxyfinder.go:97: [1] GET http://google.com/ via "DIRECT" (not connected to PAC server)
2025/07/26 16:14:45.387989 proxyfinder.go:97: [1] GET http://google.com/ via "DIRECT" (not connected to PAC server)
```

I also made two separate requests to google.com (one via 127.0.0.1 and the other via ::1) and they both get a correlation id of [1]. This shouldn't happen; each request should get a unique id.

To fix this, we can go back to creating a single server, but create multiple listeners:

```suggestion
	s := createServer(*port, *pacurl, a)
	for _, host := range hosts {
		for _, network := range networks(host) {
			go func(network string) {
				address := net.JoinHostPort(host, strconv.Itoa(*port))
				l, err := net.Listen(network, address)
```

Inside the `createServer()` function, you'll also need to delete the `host` argument, and remove the line that sets `Addr: net.JoinHostPort(host, strconv.Itoa(port)),` (and the comment above it).

main.go
samuong, line 1: If you're happy with your changes being licensed under the Apache 2.0 license, can you update the first line of this file (the copyright line) to contain 2025, and also add yourself to the AUTHORS file?

samuong (commented):
Thanks! This is looking really good, I just noticed a minor bug which hopefully I've explained clearly below (if not, please let me know).

LucasSnatiago:
Can you check if it's working now. On my computer is giving me this result:
```
go run . -l ::1 -l 127.0.0.1
2025/07/26 08:47:39.466417 main.go:90: Credentials not found, disabling proxy auth: cannot get user secret from keyring: secret not found in keyring
2025/07/26 08:47:39.470785 pacfetcher.go:94: No PAC URL specified or detected; all requests will be made directly
2025/07/26 08:47:39.471036 main.go:115: Listening on tcp4 127.0.0.1:3128
2025/07/26 08:47:39.471088 main.go:115: Listening on tcp6 [::1]:3128
2025/07/26 08:48:08.479297 proxyfinder.go:97: [1] GET http://google.com/ via "DIRECT" (not connected to PAC server)
2025/07/26 08:48:11.685076 proxyfinder.go:97: [2] GET http://google.com/ via "DIRECT" (not connected to PAC server)
2025/07/26 08:48:13.186504 proxyfinder.go:97: [3] GET http://google.com/ via "DIRECT" (not connected to PAC server)
2025/07/26 08:48:25.316086 proxyfinder.go:97: [4] CONNECT //google.com:443 via "DIRECT" (not connected to PAC server)
```
I did 4 requests to google and every single one received only one id so it appears to be fixed.
I'm not at my company this weekend to check if its working through the proxy. Monday I can check it.

samuong (approved):
LGTM, thanks!

PR #153, opened 2025-07-07
feat: experimental support for socks5

LucasSnatiago:
Adds an experimental support for socks5. 
It basically makes a tunneling from socks 5 to HTTP.

I made this pr on top of #152.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,11 +16,13 @@ require (
 
 require (
 	github.com/alessio/shellescape v1.4.1 // indirect
+	github.com/armon/go-socks5 v0.0.0-20160902184237-e75332964ef5 // indirect
 	github.com/danieljoos/wincred v1.2.0 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/godbus/dbus/v5 v5.1.0 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	golang.org/x/crypto v0.24.0 // indirect
+	golang.org/x/net v0.21.0 // indirect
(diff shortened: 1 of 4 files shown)


samuong:
Just a heads up that I haven't forgotten about this, I'm still looking into it. I don't have a lot of experience with SOCKS proxies, so I'm still trying to understand everything.

In particular, is there any reason you'd prefer to use a SOCKS proxy rather than a regular HTTP proxy?

My understanding is that SOCKS proxies are nice because they work with UDP and non-HTTP protocols. But Alpaca is only ever going to support HTTP since we're doing auth through HTTP headers. Please let me know if I've missed anything.

samuong:
Ok, there's a few things I feel we need to figure out before we can enable socks support for users:

Unless things have changed since [this article](https://eli.thegreenplace.net/2022/go-and-proxy-servers-part-3-socks-proxies/) was written, the github.com/armon/go-socks5 won't support UDP.

Another is how to deal with upstream socks proxies. This PR will add a port to alpaca for socks clients to connect to, but there's also the other side: if the PAC function returns a socks proxy, alpaca needs to connect to that proxy and forward traffic there. This also opens up the questions of whether and how to support authentication to these upstream proxies.

I think it would make sense to keep this in a separate branch for now, but if you want to work together to figure these things out and put together a plan then let me know.

LucasSnatiago:
I studied SOCKS proxy a bit more, and it’s indeed difficult to maintain. To keep this project’s scope focused, it might be a good idea not to support it. If you agree, feel free to close this issue.

PR #156, opened 2025-07-28
Fix compile errors on Darwin

samuong:


samuong:
A todo for later:

I've noticed that on all platforms except Windows, `pacFinder.pacChanged()` will call `pacFinder.findPACURL()`. But even on Windows, `pacFinder.findPACURL()` could be modified to return both the URL as well as a bool to indicated whether that URL changed. This would both simplify the implementation, and remove some extra checks that Alpaca performs unnecessarily.

We could probably also remove the error return value, and make it so that it is _never_ a fatal error if the function fails. Implementations could log any errors they encounter, if it is useful to do so.

PR #157, opened 2025-10-22
Feat: Add support for data URL encoded PACs

m04f:
This adds support for reading pac files from [data urls](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Schemes/data)

### DIFF
--- a/pacfetcher.go
+++ b/pacfetcher.go
@@ -30,6 +30,11 @@ import (
 // The maximum size (in bytes) allowed for a PAC script. At 1 MB, this matches the limit in Chrome.
 const maxResponseBytes = 1 * 1024 * 1024
 
+// The maximum size (in bytes) allowed for a data URL.
+// Chromium and Firefox limit data URLs to 512MB.
+// See https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Schemes/data#length_limitations
+const maxDataURLLength = 512 * 1024 * 1024
+
 // The time to wait before retrying a failed PAC download. This is similar to Chrome's delay:
 // https://cs.chromium.org/chromium/src/net/proxy_resolution/proxy_resolution_service.cc?l=96&rcl=3db5f65968c3ecab3932c1ff7367ad28834f9502
 var delayAfterFailedDownload = 2 * time.Second
@@ -95,6 +100,11 @@ func decodeDataURL(uri string) ([]byte, error) {
 	if parsedURL.Scheme != "data" {
 		return nil, nil
 	}
+
+	if len(uri) > maxDataURLLength {
+		return nil, fmt.Errorf("Error parsing data URL: PAC JS is too big (limit is %d bytes)", maxDataURLLength)
+	}
+
 	metadata, data, ok := strings.Cut(parsedURL.Opaque, ",")
 	if !ok {
 		return nil, fmt.Errorf("Error parsing data URL: Invalid Format")
@@ -105,18 +115,12 @@ func decodeDataURL(uri string) ([]byte, error) {
 		if err != nil {
 			return nil, fmt.Errorf("Error decoding base64 data URL: %w", err)
 		}
-		if len(bytes) > maxResponseBytes {
-			return nil, fmt.Errorf("Error parsing base64 data URL: PAC JS is too big (limit is %d bytes)", maxResponseBytes)
-		}
 		return bytes, nil
 	}
 
 	decoded, err := url.QueryUnescape(data)
 	if err != nil {
 		return nil, fmt.Errorf("Error parsing data URL: %w", err)
-	}
-	if len(decoded) > maxResponseBytes {
-		return nil, fmt.Errorf("Error parsing data URL: PAC JS is too big (limit is %d bytes)", maxResponseBytes)
 	}
 	return []byte(decoded), nil
 }
--- a/pacfetcher.go
+++ b/pacfetcher.go
@@ -40,10 +40,10 @@ const maxDataURLLength = 512 * 1024 * 1024
 var delayAfterFailedDownload = 2 * time.Second
 
 type pacFetcher struct {
-	pacFinder  *pacFinder
-	monitor    netMonitor
-	client     *http.Client
-	connected  bool
+	pacFinder *pacFinder
+	monitor   netMonitor
+	client    *http.Client
+	connected bool
 	//cache  []byte
 	//modified time.Time
 	//fetched time.Time
@@ -70,9 +70,9 @@ func newPACFetcher(pacurl string) *pacFetcher {
 		client.Transport = &http.Transport{Proxy: nil}
 	}
 	return &pacFetcher{
-		pacFinder:  newPacFinder(pacurl),
-		monitor:    newNetMonitor(),
-		client:     client,
+		pacFinder: newPacFinder(pacurl),
+		monitor:   newNetMonitor(),
+		client:    client,
 	}
 }
 
@@ -94,7 +94,7 @@ func decodeDataURL(uri string) ([]byte, error) {
 	parsedURL, err := url.Parse(uri)
 
 	if err != nil {
-		return nil, fmt.Errorf("Error parsing pac url: %w", err)
+		return nil, fmt.Errorf("error parsing pac url: %w", err)
 	}
 
 	if parsedURL.Scheme != "data" {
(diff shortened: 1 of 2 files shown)


pacfetcher.go
```diff
@@ -80,6 +83,46 @@ func requireOK(resp *http.Response, err error) (*http.Response, error) {
...
+// It supports both base64 and URL-encoded data, and enforces the maxResponseBytes size limit.
+// See https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs for details.
+func decodeDataURL(uri string) ([]byte, error) {
+	parsed_url, err := url.Parse(uri)
+
+	if err != nil {
+		return nil, fmt.Errorf("Error parsing pac url: %w", err)
+	}
+
+	if parsed_url.Scheme != "data" {
```
samuong: we should call this `parsedURL` as per https://go.dev/doc/effective_go#mixed-caps

```suggestion
	parsedURL, err := url.Parse(uri)

	if err != nil {
		return nil, fmt.Errorf("Error parsing pac url: %w", err)
	}

	if parsedURL.Scheme != "data" {
```

pacfetcher.go
```diff
@@ -80,6 +83,46 @@ func requireOK(resp *http.Response, err error) (*http.Response, error) {
...
+		return nil, nil
+	}
+	parts := strings.SplitN(parsed_url.Opaque, ",", 2)
+	if len(parts) < 2 {
+		return nil, fmt.Errorf("Error parsing data URL: Invalid Format")
+	}
+
+	isBase64 := slices.Contains(strings.Split(parts[0], ";"), "base64")
+	if isBase64 {
```
samuong: i think it would be nicer if you used [strings.Cut()](https://pkg.go.dev/strings#Cut) and [strings.HasSuffix()](https://pkg.go.dev/strings#HasSuffix) so we don't have to deal with slices and indexes, what do you think?

```suggestion
	metadata, data, ok := strings.Cut(parsedURL.Opaque, ",")
	if !ok {
		return nil, fmt.Errorf("Error parsing data URL: Invalid Format")
	}

    if strings.HasSuffix(metadata, ";base64") {  // ignore media type
```

if you use slices.Contains(), it'll check for the substring `base64` on both sides of the `;`?

pacfetcher.go
```diff
@@ -80,6 +83,46 @@ func requireOK(resp *http.Response, err error) (*http.Response, error) {
...
+		bytes, err := base64.StdEncoding.DecodeString(parts[1])
+		if err != nil {
+			return nil, fmt.Errorf("Error parsing base64 data URL: %w", err)
```
samuong: i'm nitpicking here, but it should probably say "decoding" instead of "parsing"

pacfetcher_test.go
```diff
@@ -129,3 +129,24 @@ func TestPacFromFilesystem(t *testing.T) {
 	assert.Equal(t, content, pf.download())
 	assert.True(t, pf.isConnected())
 }
+
+func TestDecodeDataURL_Base64(t *testing.T) {
+	uri := "data:application/x-ns-proxy-autoconfig;base64,ZnVuY3Rpb24gRmluZFByb3h5Rm9yVVJMKHVybCwgaG9zdCkgewogIHJldHVybiAiUFJPWFkgcHJveHk6ODA4MCI7Cn0K"
+	want := []byte("function FindProxyForURL(url, host) {\n  return \"PROXY proxy:8080\";\n}\n")
+	pf := newPACFetcher(uri)
+	assert.Equal(t, want, pf.download())
```
samuong: i find the output of assert.Equal() to be really hard to read when the want/got values are of type `[]byte` (rather than `string) and the assertion fails. can you convert the arguments to strings before doing the assertion?

```suggestion
	want := "function FindProxyForURL(url, host) {\n  return \"PROXY proxy:8080\";\n}\n"
	pf := newPACFetcher(uri)
	assert.Equal(t, want, string(pf.download()))
```

pacfetcher_test.go
```diff
@@ -129,3 +129,24 @@ func TestPacFromFilesystem(t *testing.T) {
...
+	want := "function FindProxyForURL(url, host) {\n  return \"PROXY proxy:8080\";\n}\n"
+	pf := newPACFetcher(uri)
+	assert.Equal(t, want, string(pf.download()))
+}
+
+func TestDecodeDataURL_NonDataScheme(t *testing.T) {
+	uri := "http://example.com"
+	got, err := decodeDataURL(uri)
+	assert.Nil(t, got)
+	assert.Nil(t, err)
```
samuong:
```suggestion
	assert.NoError(t, err)
```

pacfetcher_test.go
```diff
@@ -129,3 +129,24 @@ func TestPacFromFilesystem(t *testing.T) {
...
+	pf := newPACFetcher(uri)
+	assert.Equal(t, want, pf.download())
+}
+
+func TestDecodeDataURL_URLEncoded(t *testing.T) {
+	uri := "data:,function%20FindProxyForURL(url%2C%20host)%20%7B%0A%20%20return%20%22PROXY%20proxy%3A8080%22%3B%0A%7D%0A"
+	want := "function FindProxyForURL(url, host) {\n  return \"PROXY proxy:8080\";\n}\n"
+	pf := newPACFetcher(uri)
+	assert.Equal(t, want, string(pf.download()))
+}
```
samuong: can you convert these into [table driven tests](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests)? i haven't always been consistent with this in the alpaca codebase, but you can see some examples in `requestlogger_test.go` and `pacrunner_test.go`

pacfetcher.go
```diff
@@ -80,6 +83,46 @@ func requireOK(resp *http.Response, err error) (*http.Response, error) {
...
+		}
+		if len(bytes) > maxResponseBytes {
+			return nil, fmt.Errorf("Error parsing base64 data URL: PAC JS is too big (limit is %d bytes)", maxResponseBytes)
+		}
+		return bytes, nil
+	}
+
+	decoded, err := url.QueryUnescape(parts[1])
```
samuong: after reading the mdn link, it seems to me that percent decoding should happen before base64 decoding, but i'm not sure if i'm reading it incorrectly, what do you think?

if we do do percent decoding before base64 decoding, should we use [url.PathUnescape()](https://pkg.go.dev/net/url#PathUnescape) here instead? the `+` character is part of the standard base64 alphabet

> PathUnescape is identical to [QueryUnescape](https://pkg.go.dev/net/url#QueryUnescape) except that it does not unescape '+' to ' ' (space).

pacfetcher.go
```diff
@@ -80,6 +83,46 @@ func requireOK(resp *http.Response, err error) (*http.Response, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("Error parsing data URL: %w", err)
+	}
+	if len(decoded) > maxResponseBytes {
+		return nil, fmt.Errorf("Error parsing data URL: PAC JS is too big (limit is %d bytes)", maxResponseBytes)
+	}
```
samuong: i'd suggest doing this check once, before we try do decode the data: url. we should probably use a different limit than `maxResponseBytes`. from your mdn link:

> Browsers are not required to support any particular maximum length of data. Chromium and Firefox limit data URLs to 512MB, and Safari (WebKit) limits them to 2048MB. Note that Firefox 97 increased the limit from 256KB to 32MB, and [Firefox 136 increased it to 512MB](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/136).

pacfetcher_test.go
samuong:
```suggestion
// Copyright 2019, 2021, 2022, 2025 The Alpaca Authors
```

samuong (commented):
thanks for contributing! i've left a few comments below

pacfetcher.go
samuong: OK according to https://datatracker.ietf.org/doc/html/rfc2397:

> The appearance of ";base64" means that the data
   is encoded as base64. Without ";base64", the data (as a sequence of
   octets) is represented using ASCII encoding for octets inside the
   range of safe URL characters and using the standard %xx hex encoding
   of URLs for octets outside that range.

So you're correct to do this outside of the branch where we base64-decode the data, which is good. But I think if we use url.QueryEscape() then it'll convert any `+` characters to ` `, so for example if we have `data:,2+2=5` as our input, we'll get `2 2=5` as our output, which is incorrect - so we should use url.PathEscape() instead?

m04f:
Thanks for your time; I've fixed these issues.
Sorry for the mistakes, I haven't written much go code yet.

samuong:
Not at all, this looks great, thanks again for the pull request! LGTM, I'll merge this one shortly.

PR #158, opened 2025-10-25
feat: Add SOCKS5 proxy support

m04f:
This pull request adds very basic support for SOCKS5 proxies to Alpaca.

  The following changes are included:

  - The connectViaProxy function in proxy.go now handles SOCKS5 proxies by using the proxy.SOCKS5 dialer.
  -  The findProxyForRequest function in proxyfinder.go now recognizes "SOCKS5" as a valid proxy scheme.
  -  New tests have been added to proxy_test.go to verify the SOCKS5 proxy functionality.
  -  A new test case has been added to proxyfinder_test.go to ensure that the proxy finder can correctly parse SOCKS5 proxy configurations.
  
  The changes have been tested and are working as expected.

  Note: This is a very basic implementation and has the following limitations:

  - It does not support authentication.
  - It only supports SOCKS5 (I'm not familiar with these protocols so I don't know how hard would it be to support other versions).


m04f:
hi @samuong, will you be able to review this anytime soon?

samuong:
hi @m04f, thanks for contributing! sorry i'm so late in replying to this, but i'll try to find some time over the holidays to go over this. in the meantime, i'm curious to learn a bit more about what your intended use case for socks is?

fyi there was some previous discussion in #153 and one thing we noted was that socks is useful for udp tunnelling, but the library that was being used won't support udp (see https://github.com/samuong/alpaca/pull/153#issuecomment-3116170664 for a link)

i noticed in the readme for https://github.com/things-go/go-socks5 that it references "original armon's [go-sock5](https://github.com/armon/go-socks5) library", do you know if this library has the same limitation?

m04f:
I use SSH SOCKS5 forwarding to access a jump server, so I need Alpaca to support SOCKS5 in PAC files. Unlike previous pr (which tried to add a SOCKS server), this PR just adds basic client support.

Also, things-go/go-socks5 is only used in the test suite, not in the main binary.

sorry for replying this late

samuong:
Thanks for clarifying that things-go/go-socks5 is only used for testing. I think this looks reasonable, but since there are still some limitations we should probably put this behind a command-line flag like `enable-socks`. If you can add that, I'm happy to merge.

m04f:
sorry for the delay, I've added a new `-enable-socks` flag to opt-in socks5

samuong (approved):
LGTM, and thanks for working on this!

Issue #160, opened 2025-12-17
Alpaca Icon


Enelass:
It's been bugging for a long time...
@samuong can you please make an official icon for Alpaca?
For instance something like...
<img width="1536" height="1024" alt="Image" src="https://github.com/user-attachments/assets/f96d1bc2-d7c3-49ec-93df-9d4525ac0a25" />

samuong:
Looks great! Out of these, I think I like the one on the bottom-left the most. Do you own the copyright to these, and if so are you happy to create a PR to add it to the README?

Enelass:
I do not own the copyright, you do now: Merry Christmas 🎁 🎄 !
Courtesy of AI

https://github.com/samuong/alpaca/pull/161

PR #161, opened 2025-12-24
Update readme with images

Enelass:
Adds a visual example to the README, reflecting the idea that a picture is worth a thousand words.

As per https://github.com/samuong/alpaca/issues/160#issuecomment-3688382003


samuong (approved):
lgtm, thanks!

PR #168, opened 2026-02-14
Add MacOS Kerberos support and HTTP Basic Authentication for proxies

omebarki:
In corporate environment we have several HTTP proxies with various authentication protocols.
I added support for basic auth and Kerberos for Mac OS and the ability to make them all working simultaneously


mkilchhofer:
I am compiled this PR some days ago and use the macOS Kerberos feature since then.

It seems to work well 😎 .
@omebarki thank you for this effort, lets hope that you get a review soon (by @samuong )

main.go
```diff
@@ -58,11 +59,19 @@ func main() {
 	port := flag.Int("p", 3128, "port number to listen on")
 	pacurl := flag.String("C", "", "url of proxy auto-config (pac) file")
 	domain := flag.String("d", "", "domain of the proxy account (for NTLM auth)")
-	username := flag.String("u", whoAmI(), "username of the proxy account (for NTLM auth)")
+	username := flag.String("u", whoAmI(), "username for proxy auth (NTLM)")
+	basicCreds := flag.String("b", "", "login:password for basic proxy auth")
```
samuong: For NTLM credentials, we get it from the `NTLM_CREDENTIALS` env var. We should get basic credentials from `BASIC_CREDENTIALS`. This avoids the password getting written to the shell history or made available to other processes that look.

main.go
```diff
@@ -101,9 +117,27 @@ func main() {
 		os.Exit(0)
 	}
 
+	// Build auth chain: Kerberos → Basic → NTLM.
```
samuong: Chrome's auth chain goes Negotiate -> NTLM -> Digest -> Basic. Let's do the same order (we can skip digest)?

> The Basic scheme has the lowest score because it sends the username/password unencrypted to the server or proxy.

https://www.chromium.org/developers/design-documents/http-authentication/

multiauth.go
```diff
@@ -0,0 +1,99 @@
...
+import (
+	"fmt"
+	"log"
+	"net/http"
+	"net/url"
+	"sync"
+)
+
+// multiAuthenticator tries multiple authentication methods in order and caches
+// which method works for each proxy host to avoid redundant retries.
```
samuong, line 27: This is a cool idea but the proxy server will send a Proxy-Authenticate response header, and I think we need to follow that: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Proxy-Authenticate

samuong (commented):
Thanks @omebarki for contributing, and @mkilchhofer for testing it - it's great to hear this is working for multiple people.

Sorry for taking so long to review, I've been busy with other things, but also the main blocker for me reviewing this was that I'm not able to test this myself. Right now, I neither have a Mac to run this on, nor a Kerberos server to authenticate against. I can see this continuing to be a problem for me and potentially others in the future. Ideally, this PR would include an end-to-end test (similar [main_test.go](https://github.com/samuong/alpaca/blob/5bc3447cc77e6c385e183d9514a0f13fca011824/main_test.go) which tests alpaca against squid) that tests alpaca against some open source implementation of Kerberos. I appreciate this might be very difficult to do, so you can leave it out if it's not practical, but maybe you could ask Claude to give it a go and see what it comes up with? :)

In addition to testing against a Kerberos server, it would also be good to have kerberos_darwin_test.go (unless there's too much test user setup involved), multiauth_test.go (but see my comment about following the Proxy-Authenticate header) and basicauth_test.go. Also, do we need to prompt harder to do this in CLAUDE.md?

multiauth.go
```diff
@@ -0,0 +1,151 @@
...
+	if proxyHost != "" {
+		m.mu.RLock()
+		cached, ok := m.cache[proxyHost]
+		m.mu.RUnlock()
+		if ok {
+			return cached.do(req, rt)
+		}
+	}
+
+	// Make a probe request to discover which auth schemes the proxy supports
```
samuong, line 76: I think your PR will make alpaca work like this:

1. in proxy.go (either in proxyRequest or connectViaProxy, depending on the http method), we'll send a request and we might get a 407 proxy auth required response, in which case we'll enter multiAuthenticator.do()
2. first we'll try the cache to see what worked previously, and the cache is keyed on the proxy host name
3. then we'll filter the auth schemes and try what we can

So there are a couple of issues here:

1. the probe request is redundant, and in order to avoid this roundtrip, we should take the proxy-authenticate response header from proxy.go and feed it to multiAuthenticator
2. the way that this is currently implemented, if the proxy responds with a different set of auth schemes, we might use the a stale cache entry (e.g. maybe whoever is running the proxy has upgraded it and now it supports a different set of auth schemes)

I think the cache is nice to have but if you're not seeing performance issues then I'm also happy to leave it out, especially since we're also removing the "probe" request. Given the server is telling Alpaca what auth schemes it supports, I would expect the first one to work in most cases anyway?

samuong (commented):
Thanks, this is looking really good, but I still see some issues in the multiAuthenticator, see my comment below.

samuong (approved):
Hi @omebarki, this Proxy-Authenticate header handling issue is the only one left in this PR, but to keep things moving I think it's ok if we do that fix that in a separate PR. I just fixed a merge conflict (I merged your other PR earlier today) so I'll approve and merge this. Thanks again for your contributions!

samuong:
Also, sorry I did a terrible job of fixing merge conflicts, this has been fixed in #177. I'll avoid resolving merge conflicts in the web UI from now on :sweat_smile:

smithjw:
Thanks for getting started @omebarki and for merging it in @samuong. I'd started working on some of the feedback comments with these changes forming the base but glad those can just be included in a clean PR I'll raise in the next couple days.

I'll follow up with a PR for Windows support soon after :)

PR #179, opened 2026-05-30
Remove kerberos wait (-w) flag

samuong:


samuong:
@omebarki and @smithjw do you have any opinions on this?

This was added in #168 but #178 modified this so that Alpaca will check for the kerberos ticket on a 407, rather than at startup, so I feel like it is now unnecessary. I'd like to err on the side of not adding new flags during releases, to minimise the maintenance burden.

If there are no objections by Monday morning (AEST) I'll merge this and push out a release. If you still think the flag is useful after that, please still let me know, and we can add it back if needed.

smithjw:
Yeah, perfect. I didn't think this flag would be needed any further and happy you cleaned this up with the new release :)

PR #182, opened 2026-06-26
Eliminate per-connection 407 probe for known-auth proxies

### LINKED ISSUE #181: Reduce redundant 407 round-trips on CONNECT tunnels for known-auth proxies
### Observed behaviour
Every HTTPS request through an NTLM-authenticated proxy costs three round-trips:
1. Unauthenticated CONNECT → 407
2. CONNECT + NTLM Type 1 → 407 + Type 2 challenge
3. CONNECT + NTLM Type 3 → 200

Step 1 can be redundant on every connection after the first attempt — IF the proxy's auth requirement doesn't change between requests.

### Root cause
[connectViaProxy() in proxy.go](https://github.com/samuong/alpaca/blob/master/proxy.go#L348) unconditionally fires an unauthenticated RoundTrip before checking for a 407.
There is no mechanism to remember that a given proxy host has previously demanded authentication.

### Impact/ROI
Not a bug, but a feature request for users where they are hitting a known host/proxy which requires Auth, to eliminate one round-trip per CONNECT tunnel after warm-up, reducing latency and proxy log noise. This is particularly noticeable when a tool makes frequent short-lived HTTPS requests to the same endpoint. eg Claude Code (multiple sessions + multiple agents) hitting the same inference endpoint via a corporate mandated proxy.

Image

### Proposed Options
@samuong, wanted your thoughts on this, before looking at implementing a fix, in case there were any other paths, you may have in mind already or if you don't feel this feature is desired.

#### Option 1 - In memory cache
Add a per-session cache (keyed on proxy host) that stores the auth schemes advertised in the first 407 response.
On subsequent CONNECT tunnels to the same host, skip the unauthenticated probe and call retryConnectWithAuth directly with the cached schemes. The NTLM Type 1/2/3 handshake is unchanged.
  - Cache is only populated on a 407 — proxies that don't require auth are never cached and continue to receive the
  unauthenticated probe
  - RFC 7235 §4.3 explicitly permits this: "usually, but not necessarily, after receiving a 407"
  - The cache is conservative: it requires observed evidence (a prior 407 from that host in this session) before skipping the probe

#### Option 2 - Static host list
Static host list via env var — user declares known-auth proxy hosts upfront, skipping the probe from the first connection.
This follows the same pattern as the existing ALPACA_PROXY_AUTH_ALLOWLIST.
In connectViaProxy, if the proxy host is in the list, skip the unauthenticated probe and call retryConnectWithAuth directly with a default scheme set (["Negotiate", "NTLM", "Basic"]) — auth.pick() intersects that with what's actually configured anyway.
- Simpler code, follows existing ALPACA_PROXY_AUTH_ALLOWLIST pattern, but requires user configuration.

rurouni88:
### Summary

Adds a per-session auth cache (keyed on proxy host) to ProxyHandler. After the first CONNECT tunnel to a proxy receives a 407, the advertised schemes are cached. Subsequent tunnels skip the unauthenticated probe and go straight to authentication.
Net result: One 407 on session warm-up, zero on every subsequent CONNECT to the same proxy host. A stale entry self-heals in one extra round-trip.

### What's added
- proxyAuthInfo struct — holds the schemes advertised in a proxy's 407 response.
- authCache *sync.Map on ProxyHandler — per-session cache keyed on proxy host, initialised in NewProxyHandler.
- Phase 1 / Phase 2 split in connectViaProxy — Phase 1 (cache hit) skips the unauthenticated probe and calls retryConnectWithAuth directly with cached schemes. Phase 2 (cache miss) is the original behaviour; on 407 it populates the cache before retrying.
- Stale-entry eviction — a cache hit that returns 407 evicts the entry, runs a cold probe on a fresh transport, and repopulates on successful re-auth.
- onPACUpdate func() hook on ProxyFinder — called after every successful PAC re-download; wired in main.go to flush the cache so network topology changes (VPN, Wi-Fi switch) clear stale entries automatically. Kept it simple, because this could have been a rabbit hole. As that old adage goes:
> "There are only two hard things in Computer Science: _cache invalidation_, naming things, and off-by-one errors." - Leon Bambrick
- Seven new tests — five TestConnectAuthCache_* in proxy_test.go covering cache population, probe suppression, stale eviction (persistent 407), stale eviction then success, and the two-pass probe-count assertion requested in #181; two TestProxyFinder_* confirming the PAC flush callback fires on re-download and not on failure.

### What's changed (with why)
- connectViaProxy gains a 4th authCache *sync.Map parameter. Two existing call sites in multiauth_integration_test.go updated to pass &sync.Map{}.
- req.Header.Del("Proxy-Authorization") before the cold probe in the stale-eviction path. After retryConnectWithAuth runs, the request carries a residual auth header — clearing it    ensures the cold probe is genuinely unauthenticated.
- RFC 7235 §4.3 updated to RFC 9110 §11.3 throughout.

### How to validate
Manual: with a live NTLM proxy, the first CONNECT logs a 407 handshake; all subsequent CONNECTs go straight to Attempting NTLM authentication with no preceding 407. Verified on Windows against a corporate NTLM proxy. Screencap with redacted info below)
image

### Notes for review
- Plain HTTP path (proxyRequest) is unchanged — NTLM persistence there is handled by Go's transport layer.
- No LRU cap — as discussed in #181, the cache is keyed on PAC-sourced proxy hosts (trusted, bounded), not arbitrary web server hosts. Firefox has no cap either.
- onPACUpdate callback design was discussed and confirmed in #181. Construction order in main.go (ProxyHandler before ProxyFinder) is load-bearing — noted in a comment.
- Fixes #181

### DIFF
--- a/main.go
+++ b/main.go
@@ -208,12 +208,7 @@ func createServer(port int, pacurl string, auth *authChain, enableSocks bool) *h
 			blockProxy(host)
 		}
 	})
-	onPACUpdate := func() {
-		proxyHandler.authCache.Range(func(k, _ any) bool {
-			proxyHandler.authCache.Delete(k)
-			return true
-		})
-	}
+	onPACUpdate := proxyHandler.authCache.Clear
 	proxyFinder := NewProxyFinder(pacurl, pacWrapper, enableSocks, onPACUpdate)
 	blockProxy = proxyFinder.blockProxy
 	mux := http.NewServeMux()
--- a/main.go
+++ b/main.go
@@ -200,8 +200,22 @@ func main() {
 
 func createServer(port int, pacurl string, auth *authChain, enableSocks bool) *http.Server {
 	pacWrapper := NewPACWrapper(PACData{Port: port})
-	proxyFinder := NewProxyFinder(pacurl, pacWrapper, enableSocks)
-	proxyHandler := NewProxyHandler(auth, getProxyFromContext, proxyFinder.blockProxy)
+	// Construction order is load-bearing: authCache must exist before
+	// onPACUpdate is captured, and blockProxy before proxyFinder is used.
+	var blockProxy func(string)
+	proxyHandler := NewProxyHandler(auth, getProxyFromContext, func(host string) {
+		if blockProxy != nil {
+			blockProxy(host)
+		}
+	})
+	onPACUpdate := func() {
+		proxyHandler.authCache.Range(func(k, _ any) bool {
+			proxyHandler.authCache.Delete(k)
+			return true
+		})
+	}
+	proxyFinder := NewProxyFinder(pacurl, pacWrapper, enableSocks, onPACUpdate)
+	blockProxy = proxyFinder.blockProxy
 	mux := http.NewServeMux()
 	pacWrapper.SetupHandlers(mux)
 
--- a/main.go
+++ b/main.go
@@ -200,8 +200,9 @@ func main() {
 
 func createServer(port int, pacurl string, auth *authChain, enableSocks bool) *http.Server {
 	pacWrapper := NewPACWrapper(PACData{Port: port})
-	// Construction order is load-bearing: authCache must exist before
-	// onPACUpdate is captured, and blockProxy before proxyFinder is used.
+	// Note that proxyHandler and proxyFinder are mutually dependent:
+	// proxyHandler calls proxyFinder.blockProxy(), and proxyFinder calls
+	// proxyHandler.authCache.Clear().
 	var blockProxy func(string)
 	proxyHandler := NewProxyHandler(auth, getProxyFromContext, func(host string) {
 		if blockProxy != nil {
--- a/proxy.go
+++ b/proxy.go
@@ -415,6 +415,7 @@ func connectViaProxy(req *http.Request, proxyURL *url.URL, auth *authChain,
 				log.Printf("[%d] Got %q response, retrying with auth", id, resp2.Status)
 				schemes := parseProxyAuthenticateSchemes(resp2.Header)
 				_ = resp2.Body.Close()
+				authCache.Store(proxyURL.Host, proxyAuthInfo{schemes: schemes})
 				authResp2, err := retryConnectWithAuth(req, proxyURL, auth, schemes, &tr2)
 				if err != nil {
 					return nil, err
--- a/proxy.go
+++ b/proxy.go
@@ -222,6 +222,7 @@ func splitChallengeNames(value string) []string {
 }
 
 type proxyAuthInfo struct {
+	// schemes are lower-cased, matching parseProxyAuthenticateSchemes output.
 	schemes []string
 }
 
--- a/proxy.go
+++ b/proxy.go
@@ -277,7 +277,7 @@ func (ph ProxyHandler) handleConnect(w http.ResponseWriter, req *http.Request) {
 	if proxyURL == nil {
 		server, err = connectDirect(req)
 	} else {
-		server, err = connectViaProxy(req, proxyURL, ph.auth, ph.authCache)
+		server, err = ph.connectViaProxy(req, proxyURL)
 		var oe *net.OpError
 		if errors.As(err, &oe) && oe.Op == "proxyconnect" {
 			log.Printf("[%d] Temporarily blocking proxy: %q", id, proxyURL.Host)
@@ -352,8 +352,40 @@ func connectDirect(req *http.Request) (net.Conn, error) {
 	return server, err
 }
 
-func connectViaProxy(req *http.Request, proxyURL *url.URL, auth *authChain,
-	authCache *sync.Map) (net.Conn, error) {
+// coldProbe dials proxyURL on tr, sends an unauthenticated CONNECT, and handles
+// the 407 response by caching the advertised schemes and retrying with auth.
+// The caller is responsible for deferring tr.Close().
+func coldProbe(req *http.Request, proxyURL *url.URL, auth *authChain,
+	tr *transport, authCache *sync.Map) (*http.Response, error) {
+	id := req.Context().Value(contextKeyID)
+	if err := tr.dial(proxyURL); err != nil {
+		log.Printf("[%d] Error dialling proxy %s: %v", id, proxyURL.Host, err)
+		return nil, err
+	}
+	req.Header.Del("Proxy-Authorization")
+	resp, err := tr.RoundTrip(req)
+	if err != nil {
+		log.Printf("[%d] Error reading CONNECT response: %v", id, err)
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusProxyAuthRequired || auth == nil {
+		return resp, nil
+	}
+	log.Printf("[%d] Got %q response, retrying with auth", id, resp.Status)
+	schemes := parseProxyAuthenticateSchemes(resp.Header)
+	_ = resp.Body.Close()
+	authResp, err := retryConnectWithAuth(req, proxyURL, auth, schemes, tr)
+	if err != nil {
+		return nil, err
+	}
+	if authResp.StatusCode != http.StatusProxyAuthRequired {
+		authCache.Store(proxyURL.Host, proxyAuthInfo{schemes: schemes})
+	}
+	log.Printf("[%d] Got %q response", id, authResp.Status)
+	return authResp, nil
+}
+
+func (ph *ProxyHandler) connectViaProxy(req *http.Request, proxyURL *url.URL) (net.Conn, error) {
 	id := req.Context().Value(contextKeyID)
 
 	// SOCKS5 short-circuit: SOCKS5 has its own authentication model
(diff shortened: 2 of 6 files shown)


main.go
```diff
@@ -200,8 +200,22 @@ func main() {
...
+		if blockProxy != nil {
+			blockProxy(host)
+		}
+	})
+	onPACUpdate := func() {
+		proxyHandler.authCache.Range(func(k, _ any) bool {
+			proxyHandler.authCache.Delete(k)
+			return true
+		})
+	}
```
samuong: is there a reason for not using sync.Map.Clear()?

```suggestion
	onPACUpdate := proxyHandler.authCache.Clear
```

proxy.go
```diff
@@ -377,26 +385,78 @@ func connectViaProxy(req *http.Request, proxyURL *url.URL, auth *authChain) (net
...
-		log.Printf("[%d] Got %q response, retrying with auth", id, resp.Status)
-		schemes := parseProxyAuthenticateSchemes(resp.Header)
-		_ = resp.Body.Close()
-		// resp is now stale; the retry helper returns a fresh one.
-		authResp, err := retryConnectWithAuth(req, proxyURL, auth, schemes, &tr)
+	activeTr := &tr // points to the transport holding the live tunnel
+
+	var resp *http.Response
+	if cached, ok := authCache.Load(proxyURL.Host); ok && auth != nil {
+		// Phase 1: cache hit — skip unauthenticated probe.
```
samuong: It says "phase 1" here, which to me suggests that two things will happen in sequence, i.e. phase 1 and then phase 2. But this is an if-else statement, with phase 2 in the else block, so we'll do either one or the other?

proxy.go
```diff
@@ -377,26 +385,78 @@ func connectViaProxy(req *http.Request, proxyURL *url.URL, auth *authChain) (net
...
+					return nil, err
+				}
+				if authResp2.StatusCode != http.StatusProxyAuthRequired {
+					authCache.Store(proxyURL.Host, proxyAuthInfo{schemes: schemes})
+				}
+				log.Printf("[%d] Got %q response", id, authResp2.Status)
+				resp = authResp2
+			} else {
+				resp = resp2
+			}
```
samuong: I'm not sure if I've missed something, but this looks largely similar to "phase 2", and I think this duplication (the "stale cache entry" case above, and the "cold probe - no cached entry" case below) is a result of handling the cache hit and miss in two separate if-else branches.

In my mind, the logic would've been something like this:

1. Try to get schemes from cache
2. If no cached schemes, do cold probe:
   a. dial, send CONNECT
   b. if 407 then parse schemes, cache them, fall through
   c. if 200 then hijack and return (no auth needed at all)
3. retryConnectWithAuth using cached or probed schemes
4. if still 407 then evict cache and return error
5. if 200 then hijack and return

Let me know if I've oversimplified this in my head, and I'm missing something?

proxy.go
```diff
@@ -345,7 +352,8 @@ func connectDirect(req *http.Request) (net.Conn, error) {
 	return server, err
 }
 
-func connectViaProxy(req *http.Request, proxyURL *url.URL, auth *authChain) (net.Conn, error) {
+func connectViaProxy(req *http.Request, proxyURL *url.URL, auth *authChain,
+	authCache *sync.Map) (net.Conn, error) {
```
samuong: this function requires both `authChain` and now `authCache`, maybe we should just turn it into a method?

```suggestion
func (ph *ProxyHandler) connectViaProxy(req *http.Request, proxyURL *url.URL) (net.Conn, error) {
```

main.go
```diff
@@ -200,8 +200,22 @@ func main() {
 
 func createServer(port int, pacurl string, auth *authChain, enableSocks bool) *http.Server {
 	pacWrapper := NewPACWrapper(PACData{Port: port})
-	proxyFinder := NewProxyFinder(pacurl, pacWrapper, enableSocks)
-	proxyHandler := NewProxyHandler(auth, getProxyFromContext, proxyFinder.blockProxy)
+	// Construction order is load-bearing: authCache must exist before
+	// onPACUpdate is captured, and blockProxy before proxyFinder is used.
```
samuong: The term "load-bearing" was a bit confusing to me, maybe we can reword it to something like this?

```suggestion
	// Note that proxyHandler and proxyFinder are mutually dependent:
	// proxyHandler calls proxyFinder.blockProxy(), and proxyFinder calls
	// proxyHandler.authCache.Clear().
```

proxy_test.go
```diff
@@ -426,3 +427,175 @@ func TestSOCKS5Proxy(t *testing.T) {
...
+	srv := httptest.NewServer(mock)
+	t.Cleanup(srv.Close)
+	return srv, mock
+}
+
+func makeConnectReq(t *testing.T) *http.Request {
+	t.Helper()
+	req, err := http.NewRequest(http.MethodConnect, "https://target.example.com:443", nil)
+	require.NoError(t, err)
+	req.Host = "target.example.com:443"
```
samuong: the call to http.NewRequest() takes a url, how come req.Host gets set again?

proxy_test.go
```diff
@@ -426,3 +427,175 @@ func TestSOCKS5Proxy(t *testing.T) {
...
+		w.Header().Set("Proxy-Authenticate", "Basic realm=\"proxy\"")
+		w.WriteHeader(http.StatusProxyAuthRequired)
+		return
+	}
+	w.WriteHeader(http.StatusOK)
+}
+
+func (m *authCacheMockProxy) counts() (bare, authed int) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
```
samuong, line 463: iiuc all the tests run the request and then the assertions sequentially, do we need a mutex at all?

if not, maybe we could simplify it by getting rid of the mutex and just accessing the count fields directly in the assertions?

proxyfinder.go
```diff
@@ -71,20 +77,26 @@ func (pf *ProxyFinder) WrapHandler(next http.Handler) http.Handler {
 
 func (pf *ProxyFinder) checkForUpdates() {
 	pf.Lock()
-	defer pf.Unlock()
```
samuong, line 74: i think it's a lot easier to reason about concurrency when we unlock the mutex with a defer statement; do you need to strictly call pf.onPACUpdate outside of the lock?

if so, i wonder if we could do something like this?

```
func (pf *ProxyFinder) checkForUpdates() {
  notify := func() {}
  // notify gets called after the mutex is unlocked
  defer func() { notify() }()
  pf.Lock()
  defer pf.Unlock()
  // ...
}
```

alternatively, split it out into two functions: one is protected by the lock, the other is responsible for calling notify()

samuong (commented):
thanks for putting this together! it's looking really good, but i have a few suggestions below

proxy_test.go
rurouni88: http.NewRequest sets req.URL but leaves req.Host empty.
But rather than hardcoding the value twice, there is a more intuitive way to fix this, which I'll make happen.

proxy_test.go
rurouni88: I'd recommend keeping it.
The sequential assertion argument only holds if you trust net/http's internal sync chain to substitute for an explicit happens-before guarantee, which isn't what the Go memory model gives you. ServeHTTP runs on its own goroutine, and the increment fires before the response write, so the read in counts() is racing against a goroutine boundary net/http never promised to fence for you.

The mutex is cheap, race-detector clean, and the conventional pattern for any httptest.Server handler touching shared state, and removing it trades a clear guarantee for an implicit one.

proxy_test.go
samuong: Oh you're right. For some reason I think I was able to remove this line and still have the tests pass, but I'm away from my computer now so I can't confirm. If you need to leave it, feel free to do so.

rurouni88: The mock proxy doesn't care about the CONNECT target, it only looks at auth headers, so tests would pass with an empty req.Host.
But our req.Host = req.URL.Host fix is recommended, because it makes the test request match what a real CONNECT looks like, and documents why the assignment is needed.

proxy.go
rurouni88: Almost. Correction on step 4: A stale 407 isn't a dead end. It evicts the cache entry and runs a cold probe on a fresh transport, which can actually succeed. Self-healing cycle, not an error return.

Good catch on the duplication. It's been extracted to a helper function.
The if-else has to stay, because a _**cache hit skips the probe entirely**_. Collapsing to a single linear flow would cold-probe every time and scrap the optimisation I'm asking for, bringing me back to Square 1 😆

rurouni88:
Screencap of traffic test after fixes from initial round of review.
image

proxyfinder.go
rurouni88: Done. Went with your defer pattern.
Had to add a nil guard when assigning notify = pf.onPACUpdate since some call sites pass nil for the callback, which caused a panic in tests.

rurouni88:
Ran a benchmark today.
Ignore the first row (legacy ProxyApp), and our SOE Alpaca is 2.0.8.
Hopefully that doesn't give away my employer 😝 

Anyways, these tests are a N=100 tests against our Corporate LLM Gateway via our Corporate Proxy.
image

proxy.go
samuong: Sorry for taking so long to reply to this, I've had a lot on my plate over the last week.

In the case of a cache hit, we'll skip step 2 which is the cold probe - this is the optimisation we want?

In step 4, we either reach it after a cold probe + retry with auth (steps 2-3), or using a cached scheme. In both cases we should evict the cache entry and return an error response to the client (eg bad gateway). So I thought it would be a dead end, let me know if I'm missing something?

Everything else looks fine to me, if we can close this one item I think we're ready to merge.

rurouni88: All good. I've been up watching World Cup. Argentina vs Egypt was wow.

Let me think about this one. I thought Step 4 should self-heal (evict + cold probe again). The scenario the self-heal was designed for is the proxy changing its auth scheme mid-session, but I just realised that scenario is already handled by the PAC flush being wired up in proxyfinder.go. When the network changes (VPN, Wi-Fi switch), the PAC re-downloads and onPACUpdate clears the cache.

The likely failure case for 407 after retryConnectWithAuth is bad credentials, not network change anyways.

TLDR; The self-heal adds complexity and an extra round-trip for an edge case that's not going to happen, whilst  not handling the common failure case well. Good spot, let me go back to the drawing board on this one.

rurouni88: Removed it completely, as extraneous logic. Been in World Cup mode, and forgot to mention that had been done.
Please review if you're happy with it. Thanks.

